PR #1, opened 2020-09-14
Geometry slicing job

floryst:
Runs geometry slicing in the background. This tries to speed up geometry slicing for large datasets.

Issues that need addressing:
- maybe use web workers instead of setTimeout
- cancel the job if the underlying dataset is removed or changed

### DIFF
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -47,6 +47,37 @@ function lpsDirToLabels(dir) {
   return label;
 }
 
+const createGeometryJob = (rep, sliceRange, view) => ({
+  name: `Geometry Slice Job for ${rep.getProxyId()}`,
+  // TODO rep and view could be deleted by this point
+  entry: (done, isCancelled) => {
+    // generate slice info
+    const axis = view.getAxis();
+    const normal = [0, 0, 0];
+    normal[axis] = 1;
+    const pos = [0, 0, 0];
+    const slices = [];
+    for (let s = sliceRange[0]; s < sliceRange[1]; s += 1) {
+      slices.push(s);
+    }
+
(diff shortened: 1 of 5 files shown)


src/components/VtkTwoView.vue
```diff
@@ -47,6 +47,37 @@ function lpsDirToLabels(dir) {
...
+    const normal = [0, 0, 0];
+    normal[axis] = 1;
+    const pos = [0, 0, 0];
+    const slices = [];
+    for (let s = sliceRange[0]; s < sliceRange[1]; s += 1) {
+      slices.push(s);
+    }
+
+    let i = 0;
+    // TODO do this in a web worker
```
timthirion, line 65: Is converting to a web worker a lot of work? If so, I would remove this comment for now and outline the work in an issue instead.

src/components/VtkTwoView.vue
```diff
@@ -86,7 +117,7 @@ export default {
     },
     sliceRange() {
       const { bounds } = this.worldOrientation;
-      return bounds.slice(this.axis * 2, (this.axis + 1) * 2);
+      return bounds.slice(this.axis * 2, (this.axis + 1) * 2).map(Math.round);
```
timthirion, line 120: Can you summarize the logic here?

src/store/jobs.js
```diff
@@ -0,0 +1,135 @@
+import Vue from 'vue';
+
+function noop() {}
+
+export const JobState = {
```
timthirion, line 5: I'm not familiar with web workers but it seems like this functionality should already be available in some library. Is it worth doing that now?

timthirion (commented):
LGTM, a few minor comments.

Issue #7, opened 2020-09-14
Painting resets 3D view


floryst:
Painting should not reset the camera in 3D view.

Issue #2, opened 2020-09-14
Default to quad view


floryst:
Quad view should be default, rather than axial primary.

PR #8, opened 2020-09-14
WIP: Various Bug Fixes

timthirion:
- [x] #2—Default to quad view
- [x] #11—visualization module should be namespaced
- [ ] #10—Add cascading delete to the UI
- [ ] #7—Painting resets the 3D view

### DIFF
--- a/src/components/ModelBrowser.vue
+++ b/src/components/ModelBrowser.vue
@@ -38,8 +38,8 @@ export default {
       vtkCache: (state) => state.data.vtkCache,
       dataIndex: (state) => state.data.index,
       modelIDs: (state) => state.data.modelIDs,
-      colorBy: (state) => state.visualization.colorBy,
     }),
+    ...mapState('visualization', ['colorBy']),
 
     models() {
       return this.modelIDs.map((id) => ({
@@ -77,7 +77,7 @@ export default {
         colorBy: { array, location },
       });
     },
-    ...mapActions(['removeData', 'setModelColorBy']),
+    ...mapActions('visualization', ['removeData', 'setModelColorBy']),
   },
 };
 </script>
--- a/src/components/ModelBrowser.vue
+++ b/src/components/ModelBrowser.vue
@@ -77,7 +77,7 @@ export default {
         colorBy: { array, location },
       });
     },
-    ...mapActions('visualization', ['removeData', 'setModelColorBy']),
+    ...mapActions('visualization', ['setModelColorBy']),
   },
 };
 </script>
--- a/src/components/PatientBrowser.vue
+++ b/src/components/PatientBrowser.vue
@@ -371,7 +371,8 @@ export default {
       return '';
     },
 
-    ...mapActions('visualization', ['selectBaseImage', 'updateScene', 'removeData']),
+    ...mapActions(['removeData']),
+    ...mapActions('visualization', ['selectBaseImage', 'updateScene']),
     ...mapActions('dicom', ['getSeriesImage']),
   },
 };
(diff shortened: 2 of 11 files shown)


timthirion:
@floryst I'm going to try to pick off some bugs from your list. We can merge whenever you feel like there's enough.

floryst:
Sounds good to me! I'll do some bug hunting to see what else I can find.

timthirion:
@floryst Can you spot check the last commit? I think I've got most of the namespacing done but would be good to know what I missed.

src/components/ModelBrowser.vue
```diff
@@ -77,7 +77,7 @@ export default {
         colorBy: { array, location },
       });
     },
-    ...mapActions(['removeData', 'setModelColorBy']),
+    ...mapActions('visualization', ['removeData', 'setModelColorBy']),
```
floryst, line 80: You can actually get rid of `removeData` for now. It turns out I don't have any code to delete models yet.

src/components/PatientBrowser.vue
```diff
@@ -371,7 +371,7 @@ export default {
       return '';
     },
 
-    ...mapActions(['selectBaseImage', 'updateScene', 'removeData']),
+    ...mapActions('visualization', ['selectBaseImage', 'updateScene', 'removeData']),
```
floryst, line 374: We want to use the top-level `removeData` action, and that should call into the visualization model's removeData, so we should use the non-namespaced version of that action.

src/components/VtkTwoView.vue
```diff
@@ -70,13 +70,14 @@ export default {
     ...mapState({
       baseImage: 'selectedBaseImage',
       baseImageExists: (state) => state.selectedBaseImage !== NO_SELECTION,
-      worldOrientation: (state) => state.visualization.worldOrientation,
-      windowing: (state) => state.visualization.windowing,
-      resizeToFit: (state) => state.visualization.resizeToFit,
-      slices: (state) => state.visualization.slices,
-      dataIndex: (state) => state.data.index,
```
floryst, line 77: Don't forget to re-add `dataIndex` to the top-level mapState.

floryst (changes requested):
It's a good first pass! We should also edit the `removeData` in `datasets.js` to call `dispatch('visualization/removeData', dataID)` before we remove annotation data.

src/components/PatientBrowser.vue
timthirion: Further evidence to clarify `removeData`!

floryst: Agreed! We should rename other uses of `removeData` to be something more specific to that module.

src/components/VolumeRendering.vue
```diff
@@ -118,10 +118,8 @@ export default {
   },
 
   computed: {
-    ...mapState({
-      baseImage: 'selectedBaseImage',
-      colorBy: (state) => state.visualization.colorBy,
-    }),
+    ...mapState({ baseImage: 'selectedBaseImage' }),
+    ...mapState('visualization', ['colorBy']),
     ...mapGetters(['baseImagePipeline', 'baseImageColorPreset']),
```
floryst, line 123: These are part of the visualization namespace (check out the getters in `src/store/visualization.js`).

floryst (changes requested):
For renaming, we can rename the other instances of `removeData` in a separate PR.

timthirion:
@floryst After all the fixups for namespacing the visualization module, most things generally appear to be functioning. Is there specific functionality I should test?

floryst:
So long as things are working, I'm good to merge this. We can fix things up later.

floryst (approved):

PR #29, opened 2021-03-02
Oriented views

floryst:
This PR rewrites the views to better handle dataset orientation.
- cameras transform to the dataset, not the other way around
- slice views now mimic an orthogonal MPR view, so there should be consistent coronal, sagittal, and axial views (assuming the  dataset orientation is correct)
  - This means the slices are along IJK, but each slice is correctly oriented in space.


src/components/VtkThreeView.vue
```diff
@@ -42,9 +42,9 @@ export default {
 
     const {
       sceneSources,
-      worldOrientation,
+      imageConfig,
```
timthirion, line 45: How about `imageParams`? `imageConfig` makes me thing of a software-level configuration for images, not image metadata.

I realize this is a big search and replace but may make things clearer overall.

timthirion (commented):
Code LGTM. One minor comment on naming.

src/components/VtkThreeView.vue
floryst: +1 I agree that `imageParams` is clearer than `imageConfig`.

timthirion:
Code LGTM but I haven't tested. Let me know if you'd like me to before merging.

timthirion (approved):

PR #31, opened 2021-03-08
Dicom multi volume series

floryst:
This adds some handling for series containing multiple volumes.

This PR is marked as a draft, as it depends on the oriented-views branch for correct viewing of the datasets.

FYI @agirault


src/io/itk-dicom/dicom.cpp
```diff
@@ -112,6 +125,89 @@ void movefile(const std::string &src, const std::string &dst) {
...
+
+bool areCosinesAlmostEqual(std::vector<double> cosines1,
+                           std::vector<double> cosines2) {
+  for (int i = 0; i <= 1; i++) {
+    std::vector<double> vec1{cosines1.at(i), cosines1.at(i + 1),
+                             cosines1.at(i + 2)};
+    std::vector<double> vec2{cosines2.at(i), cosines2.at(i + 1),
+                             cosines2.at(i + 2)};
+    double dot = dotProduct<3>(vec1, vec2);
+    if (dot < (1 - 10e-5)) {
```
agirault, line 184: should `10e-5` be an argument of `areCosinesAlmostEqual` with a default value?

src/io/itk-dicom/dicom.cpp
```diff
@@ -112,6 +125,89 @@ void movefile(const std::string &src, const std::string &dst) {
...
+      return false;
+    }
+  }
+  return true;
+}
+
+SeriesIndexType RestrictOnImageOrientation(const SeriesIndexType &index) {
```
agirault, line 191: I would add some comments, either on top or inside, to describe what this does and why. That's the core of your change.

src/io/itk-dicom/dicom.cpp
```diff
@@ -52,6 +56,7 @@ using SeriesIndexType =
     std::unordered_map<std::string,
                        std::vector<std::string>>; // seriesUID -> []filenames
```
agirault, line 67: `std::string` isn't just the seriesUID anymore right?

src/io/itk-dicom/dicom.cpp
```diff
@@ -135,24 +231,43 @@ const json import(FileNamesContainer &files) {
...
 
-  using SeriesIdContainer = std::vector<std::string>;
+  // These are not necessarily the DICOM SeriesUIDs due to possible mangling.
+  // Use only as internal keys.
   const SeriesIdContainer &gdcmSeriesUIDs = seriesFileNames->GetSeriesUIDs();
 
-  for (auto gdcmSeriesUID : gdcmSeriesUIDs) {
-    FileNamesContainer fileNames =
-        seriesFileNames->GetFileNames(gdcmSeriesUID.c_str());
+  SeriesIndexType curSeriesIndex;
```
agirault, line 248: I was confused by the use of `index`, just realized that was a `map`... no suggestion here, just a remark in case you had any thoughts on making that clearer.

src/io/itk-dicom/dicom.cpp
```diff
@@ -112,6 +125,89 @@ void movefile(const std::string &src, const std::string &dst) {
...
+double dotProduct(const std::vector<double> &vec1,
+                  const std::vector<double> &vec2) {
+  double result = 0;
+  for (int i = 0; i < N; i++) {
+    result += vec1.at(i) * vec2.at(i);
+  }
+  return result;
+}
+
+std::string encodeCosinesAsIDPart(const std::vector<double> &cosines) {
```
agirault, line 150: This method is important too since it will help define the key/id for that sub-series. Two notes:
1. It could use a comment explaining that, and summarizing the format/template of the key it produces
2. Could that be a lambda within `RestrictOnImageOrientation` since it's not used anywhere else? Or do you anticipate it will be used outside of that scope?

src/io/itk-dicom/dicom.cpp
```diff
@@ -135,24 +231,43 @@ const json import(FileNamesContainer &files) {
...
-  using SeriesIdContainer = std::vector<std::string>;
+  // These are not necessarily the DICOM SeriesUIDs due to possible mangling.
+  // Use only as internal keys.
   const SeriesIdContainer &gdcmSeriesUIDs = seriesFileNames->GetSeriesUIDs();
 
-  for (auto gdcmSeriesUID : gdcmSeriesUIDs) {
-    FileNamesContainer fileNames =
-        seriesFileNames->GetFileNames(gdcmSeriesUID.c_str());
+  SeriesIndexType curSeriesIndex;
+  for (auto seriesUID : gdcmSeriesUIDs) {
```
agirault, line 249: why renaming from `gdcmSeriesUID` to `seriesUID`? Less changes below if not doing that.

src/io/itk-dicom/dicom.cpp
```diff
@@ -135,24 +231,43 @@ const json import(FileNamesContainer &files) {
   // Does this affect series organization?
   seriesFileNames->SetLoadPrivateTags(false);
 
-  using SeriesIdContainer = std::vector<std::string>;
+  // These are not necessarily the DICOM SeriesUIDs due to possible mangling.
+  // Use only as internal keys.
   const SeriesIdContainer &gdcmSeriesUIDs = seriesFileNames->GetSeriesUIDs();
 
-  for (auto gdcmSeriesUID : gdcmSeriesUIDs) {
```
agirault, line 141: Suggestion: add a comment above `// Separate files for each series UID`

src/io/itk-dicom/dicom.cpp
agirault: Nitpick: would `Separate` be more accurate than `Restrict`?

src/store/dicom.js
```diff
@@ -149,14 +149,11 @@ export default (dependencies) => ({
...
-
-          if (gdcmSeriesUID in state.seriesIndex) {
-            seriesKey = state.seriesIndex[gdcmSeriesUID].SeriesInstanceUID;
-          } else {
-            const info = await dicomIO.readSeriesTags(gdcmSeriesUID, [
+        updatedSeries.map(async (seriesKey) => {
+          const numberOfSlices = await dicomIO.buildSeriesOrder(seriesKey);
+
+          if (!(seriesKey in state.seriesIndex)) {
+            const info = await dicomIO.readSeriesTags(seriesKey, [
```
agirault, line 156: Is the `gdcmSeriesUID in state.seriesIndex` scenario not handled anymore?

src/components/PatientBrowser.vue
```diff
@@ -93,9 +89,9 @@
                 <div class="my-2 series-list">
                   <groupable-item
                     v-for="series in getSeries(study.StudyInstanceUID)"
-                    :key="series.SeriesInstanceUID"
+                    :key="series.ITKGDCMSeriesUID"
```
agirault, line 92: Where does this key name come from? I don't see why it refers to ITK or GDCM, it should be agnostic to that. I was thinking more of `VolumeUID` to be generic? Just my 2c.

agirault (commented):
Wow, cool work. A couple of questions and suggestions for naming.

src/io/itk-dicom/base64.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+
+// from: https://en.wikibooks.org/wiki/Algorithm_Implementation/Miscellaneous/Base64#C++
+
+#define DWORD long
+#define BYTE unsigned char
+
+const static char encodeLookup[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+const static char padCharacter = '=';
+std::basic_string<char> base64Encode(std::vector<BYTE> inputBuffer)
```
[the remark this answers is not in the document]
floryst: Oh I actually stopped using that, so I can remove the base64.hpp file now.

src/io/itk-dicom/dicom.cpp
floryst: This should be a file-scoped static constant. I can also make it a default parameter, too.

src/io/itk-dicom/dicom.cpp
floryst: yeah, it's our own generated/transient UID now. I'll need to update instances of seriesUID to reflect that. Maybe objectUID?

src/io/itk-dicom/dicom.cpp
floryst: Yeah, I occasionally interchange index and map, since both can do lookups. But I can rename to "map" since its less overloaded than "index".

src/io/itk-dicom/dicom.cpp
agirault: up to you. I think having it as a default parameter is a nice balance to avoid having to update if it that specific value isn't enough for some people/cases.

src/io/itk-dicom/dicom.cpp
agirault: Ah got it... Restrict would be fine if it had Series as well (`RestrictSeriesByImageOrientation`)? Either is fine

floryst:
We have a naming issue at hand. For identifying various aspects of the hierarchy, we previously had:
- PatientID: DICOM patientID
- StudyInstanceUID: DICOM study uid
- SeriesInstanceUID: DICOM series uid

Since we now have distinct datasets coming out of a single series, we need a new UID name. Heer are suome suggestions:
- VolumeUID
- ObjectUID
- DatasetUID

Which one do you prefer? @agirault

agirault:
`Object` would not be good. On top of being too generic, it is already defined in DICOM as an instance of an SOP class/IOD. i.e.: each separate files are already defined as instances or objects.

The PR itself is called "multi-volume series", so I think `Volume` is at least an easy-to-understand option. It's also the one used by 3D Slicer. The only problem that I can think of is that it's not exactly a volume if it's only one scan, or if it's 2d images + time. So I guess, `Volume` or `Dataset`.

floryst:
Here's a other idea: imageStackUID, or stackUID. How does that sound? I think I would go with volumeUID since I think I use datasetID somewhere else...

agirault:
I think `Volume` makes sense, and the consistency with 3DSlicer is good as well.

src/io/itk-dicom/dicom.cpp
floryst: I like the lambda idea, and I'll add a comment on the format.

src/components/PatientBrowser.vue
floryst: This was from before when the application indexed on SeriesInstanceUID rather than the generated GDCM series UID. But since we are now making our own IDs, this should be replaced with references to volumes.

src/io/itk-dicom/dicom.cpp
floryst: I'll be renaming to volumeID, so there will be much bigger changes happening.

src/store/dicom.js
floryst: This will be irrelevant once we refactor to reference volumeIDs rather than seriesUIDs.

floryst:
Before this gets approved/merged, I will submit a PR to this branch containing a refactor from operating on series to operating on volumes, since the whole premise of this PR is to support handling multiple volumes within a single series.

src/components/PatientBrowser.vue
agirault: Looks like `ITKGDCMSeriesUID` is still used around in that PR?

PR #36, opened 2021-07-14
Crosshairs 

DrewLazzeriKitware:
Screen Shot 2021-07-14 at 2 03 08 PM

- [x] View coordinates often don't translate to world coordinates
- [x] Crosshairs should respond to scrolling
- [x] Make work with oriented datasets
- [x] Crosshairs should re-render when scrolling back to align slices
- [ ] What happens when you change images when crosshairs is enabled? (Set "placed" to false, update the transforms)


DrewLazzeriKitware:
@floryst I'm stuck trying to get the crosshairs to work on either of the views on the left. Only the bottom right responds to a handler. Any pointers?

src/widgets/crosshairs.js
```diff
@@ -0,0 +1,86 @@
...
+        const normal = [0, 0, 0];
+        normal[axis] = 1;
+        const origin = [0, 0, 0];
+        origin[axis] = slices['xyz'[axis]] * spacing[axis];
+
+        // plane manipulator
+        const manipulator = factory.getManipulator();
+        manipulator.setNormal(normal);
+        manipulator.setOrigin(origin);
+      }
```
floryst, line 50: We can probably ignore this bit. Rationale is based on user expected workflow:
1. Activate crosshairs. None of the views should show the crosshairs yet
2. when mousing over a view, the crosshairs should show. This is the job of onViewMouseEvent

floryst:
There's one other thing to add: when a slice is not aligned with the crosshairs, the crosshairs should not be visible for that view.

floryst:
1. Figure out which slice a view is on (store.state.visualization.slices)
2. Trigger callback when slice changes (watchStore: listens for changes on certain store state, and then triggers a callback)
3. updateVisibility: if crosshairs X/Y/Z slice matches view X/Y/Z slice, then visible=true, otherwise visible=false

DrewLazzeriKitware:
[Link to updating visibility](https://github.com/KitwareMedical/paraview-medical/blob/misc-wip/src/widgets/newPaint.js#L103)

src/widgets/crosshairs.js
```diff
@@ -0,0 +1,104 @@
+import { vec3 } from 'gl-matrix';
+import vtkCrosshairsWidget from '@/src/vtk/CrosshairsWidget';
+import {
+  onViewMouseEvent,
+  onAddedToView,
+  onWidgetStateChanged,
+  onBeforeDelete,
+} from './context';
```
DrewLazzeriKitware, line 8: @floryst do you plan to add more hooks to `./context`?

 I want to unset "placed" when another image is selected, but none of these hooks (or the others in `./context`) fire when that happens.

PR #37, opened 2021-07-19
Load zip files

DrewLazzeriKitware:


DrewLazzeriKitware:
Hmmm this package-lock.json change is suspiciously large. I'll see if that can be cut down.

package-lock.json
```diff
@@ -1,8 +1,19342 @@
 {
   "name": "paraview-medical",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
DrewLazzeriKitware, line 4: Here's the culprit - I triggered a new lockfile version by using npm 7.6.0 here. It may be now problem, it's backwards compatible: https://stackoverflow.com/questions/66150813/is-lockfileversion-2-in-package-lock-json-from-npm-7-compatible-with-older-ve

DrewLazzeriKitware: Alternatively I may downgrade npm and reinstall.

src/store/datasets.js
```diff
@@ -142,7 +170,23 @@ export const makeActions = (dependencies) => ({
...
+    const zips = files.filter((f) => getExtension(f.name) === 'zip');
     const { fileIO } = dependencies;
+    const supportedExts = ['zip']
+      .concat(Object.keys(fileIO.fileReaders))
+      .concat(Object.keys(fileIO.typeAliases));
+    if (zips.length) {
+      const nonzips = files.filter((f) => getExtension(f.name) !== 'zip');
+      const p = zips.map((file) =>
+        JSZip.loadAsync(file).then((zip) =>
+          Promise.all(zipGetSupportedFiles(zip, null, supportedExts))
```
floryst, line 183: We will have to re-think how we get the "supported files". The FileIO object uses two techniques to determine if a file is supported: file extension, and file magic. There are times when DICOM files have no extension, yet should still be processed as DICOM.

I suspect we will have to load every file in a zip and use fileIO to test to see if they are loadable. It's not performant compared to just a name extension check, but I'll assume that people loading zips only have images inside of them.

floryst:
You have a good overall approach! We will need to address the "file detection" approach, since we cannot rely on name/extension alone.

src/store/datasets.js
floryst: Let's change this to first extract all files out of the zip, and them pass them through the loader. This way we can process DICOM files that don't necessarily have the .dcm extension.

floryst:
```js
const [zips, rest] = partitionArray(fileList, isZip);
const extractedFiles = await extractAllFilesFromZips(zips) // flatten zipped files into single array
const allFiles = [...rest, ...extractedFiles]
```

PR #50, opened 2022-04-04
Switch to 'source-map' source maps

cjh1:
@floryst I was unable to get the source to load correctly in the debugger. This resolves the issue for me. It would also gives us source maps in production, at the cost of slower rebuild times.

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -28,6 +28,7 @@ module.exports = {
       .end();
   },
   configureWebpack: {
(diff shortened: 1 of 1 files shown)


floryst:
LGTM. I seem to be getting source maps just fine, but being explicit in the config can be helpful.

FYI we use semantic-release versioning, so this commit message will need to be written as a `chore` type message.

PR #51, opened 2022-04-06
feat: Switch module selection to tabs

cjh1:


cjh1:
@floryst This PR converts the drop down to tabs. I was planning on doing the UI changes in a couple of smaller PRs instead of a "big bang" approach if that works for you.

floryst:
Looking good so far!

Is it possible to make the tabs a bit denser? I'm not entirely sure how that would look, but I'm thinking stuff like "Patients & Images" should either word wrap after the ampersand, or maybe we opt to shorten that phrase entirely. Maybe reduce padding in places, too?

cjh1:
I can reduce the padding a bit, but too much and it starts to look a bit bunched up, I will experiment. I think coming up with more concise phrases would help.

cjh1:
@floryst I have made things a little denser, at least on my screen I can see 4 modules tabs. Let me know if this looks better to you.

floryst:
Density looks better! There is an odd behavior with the bg highlighting that I'm seeing. A screenshot is below.

Screen Shot 2022-04-08 at 12 35 16 PM

cjh1:
That is strange didn't notice that, thanks

cjh1:
Highlighting it fixed up ...

floryst:
LGTM! I think it would be nicer if we could hide the left/right arrows when all the way left/right, but this is fine for now. We can experiment with that in a separate PR.

PR #54, opened 2022-04-12
fix: Don't display arrows unless they are needed

cjh1:
@floryst This removes the arrows unless they are needed. I assume you think we should also recoup the space, which will require a like more work.


floryst:
Yay! Glad the change is straightforward. :)

PR #55, opened 2022-04-15
Switch to always showing arrows

cjh1:
For some reason on mobile ( when the width getis small ) the arrows are disabled by default. Override the styling to hidden them when they are disabled.


floryst:
LGTM!

PR #65, opened 2022-05-20
Prettier fixes and checks

floryst:
- Changes from `npm run prettify`
- Adds github action checks for style


floryst:
FYI @cjh1 I'm adding a code style checker for consistency, since I didn't have one before. You can run `npm run prettify` to run prettier on the entire codebase (js and ts only).

PR #74, opened 2022-06-02
feat: Move dataset spefic view config into own store

cjh1:
This moves slice and window related configuration into a separate store. This allows us to save these values on a per dataset basis.

Note: That we will need to move the zoom level into the store too, that will be done in a follow up.


floryst:
The ruler doesn't work. When clicking to place, I get the following error:

Screen Shot 2022-06-07 at 9 54 15 AM

cjh1:
Ok, will take a look

cjh1:
@floryst The error should now be resolved.

floryst:
LGTM!

PR #81, opened 2022-06-20
Add camera config to store

cjh1:


floryst:
Switching between the Liver-CT and Abdominal-MRI samples causes the view-up in the Coronal view to rotate weirdly.

cjh1:
@floryst Can you attach a screenshot? I am not resetting the center of rotate by design, but can change that is necessary.

cjh1:
OK, I an recreate it. Will take a look.

floryst:
Repro steps on the deploy preview:
1. Load Liver-CT
2. Load Abdomen-MRI
3. Switch back to the Liver-CT dataset in the Patients & Images tab

image

I also discovered that the bottom two views (1) do not have the slice and window/level info, which makes me suspect some state isn't fully propagating to them.

cjh1:
Strange its only when you load a DICOM image ...

floryst:
AFAIK I think the Abdomen-MRI is the only dataset that has a rotated orientation. I forget if the Liver-CT one is also rotated...

cjh1:
Ok, that makes sense. Restoring the specific view up for the dataset solves the issue for me.

cjh1:
@floryst Looks like there is one more issue with the slice that I need to resolve ...

cjh1:
> @floryst Looks like there is one more issue with the slice that I need to resolve ...

Actually looks like that dataset only has two slices :-)

cjh1:
Are they hidden under the netifiy footer, this is what I am seeing? If you minimize to footer they are there or do you mean something else?

screen

floryst:
Ok, it is hidden under the netlify bar, thanks. Unfortunate that they do it that way. I had to hide the bar in order to see the text on my display.

floryst:
LGTM!

floryst:
Actually, there's an issue with the resize-to-fit functionality. When I load two datasets, resize to fit works just fine. When I pan in one view, it works as expected: resize to fit is disabled for just that view. When I switch datasets, suddenly resize-to-fit is disabled in all views.

cjh1:
@floryst So resize-to-fit is disabled if we restore a camera configuration from the store. The problem is we are receiving camera modification events for all views ( even though you only interact with one ) so resize-to-fit is disabled for all view. How is this manifesting itself as an issue in UI, new dataset are still resize to fit ? Resizing the window?

floryst:
Resize to fit should be on a per-view basis. Though in thinking about what can be done, I think tabling it for now and re-visiting the resize-to-fit logic later would be the best approach.

cjh1:
I think we probably need a way for user to reset a view, which would resize-to-fit.

floryst:
> I think we probably need a way for user to reset a view, which would resize-to-fit.

Agreed. I'll merge this first, and then fix up the resize-to-fit logic to include a "reset view" concept.

PR #82, opened 2022-06-23
Add layout to store

cjh1:
This PR ensures that the layout is added to the store, needed for state export/import. Note that there are some existing vtk.js errors when switching between the two layouts.


floryst:
LGTM!

PR #103, opened 2022-08-17
feat: migrate to itk-wasm

cjh1:
This also moves the storage of dataset files to the main thread.


src/store/datasets-dicom.ts
```diff
@@ -117,85 +118,104 @@ export const useDICOMStore = defineStore('dicom', {
...
 
-            this._updateDatabase(patient, study, volumeInfo);
-          }
-
-          // invalidate any existing volume
-          if (volumeKey in this.volumeToImageID) {
-            // buildVolume requestor uses this as a rebuild hint
-            set(this.needsRebuild, volumeKey, true);
+            // Add file entries to the file store
+            fileStore.add(volumeKey, volumeFiles);
```
floryst, line 146: In `deleteVolume`, we want to have the corresponding remove operation in the fileStore.

floryst (commented):
Looking good so far!

src/store/datasets-dicom.ts
cjh1: That is done [here](https://github.com/cjh1/paraview-medical/blob/dicom-db/src/store/datasets.ts#L259).

floryst: Oh, I see now. Thanks!

floryst:
LGTM!

PR #108, opened 2022-09-06
feat(scene): load geometry into 3D view

### LINKED ISSUE #104: Impossible to load STL and VTP files
When loading STL or VTP files (ex: ConeSources generated from Paraview), the interface freezes and the following logs show in the debug console:

```js
[Vue warn]: Error in getter for watcher "function () {

            // preventing the watch callback being call in the same execution

            if (running_1) {

                return;

            }

            try {

                running_1 = true;

                source(registerCleanup);

            }

            finally {

                running_1 = false;

            }

        }": "TypeError: curImageMetadata.value is undefined"

found in

---> <VtkTwoView> at src/components/VtkTwoView.vue
       <LayoutGrid> at src/components/LayoutGrid.vue... (1 recursive calls)
         <VMain>
           <VApp>
             <DragAndDrop> at src/components/DragAndDrop.vue
               <App> at src/components/App.vue
                 <Root> vue.runtime.esm.js:619

TypeError: curImageMetadata.value is undefined
    node_modules index.js:9051
    getter_1 vue-composition-api.mjs:1301
    get VueJS
    node_modules chunk-vendors.js:148968
    flushQueue vue-composition-api.mjs:1186
    node_modules chunk-vendors.js:148876
    VueJS 5
    fallbackFlush vue-composition-api.mjs:1193
    queueFlushJob vue-composition-api.mjs:1204
    node_modules chunk-vendors.js:148967
    VueJS 9
    setterHandler vue-composition-api.mjs:932
    set vue-composition-api.mjs:686
    useSceneBuilder useSceneBuilder.ts:44
    applyCb vue-composition-api.mjs:1372
    node_modules chunk-vendors.js:148968
    flushQueue vue-composition-api.mjs:1186
    node_modules chunk-vendors.js:148876
    VueJS 5
```

This issue can be reproduced directly on the VolView demo. I have trouble understanding where the error comes from, but I assume it's not related to the files I'm using, since I can load them in PVGlance.

floryst:
Closes #104

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -374,7 +374,9 @@ export default defineComponent({
         (loadFirstDataset || loaded.length === 1)
       ) {
         const selection = convertSuccessResultToDataSelection(loaded[0]);
-        dataStore.setPrimarySelection(selection);
+        if (selection) {
+          dataStore.setPrimarySelection(selection);
+        }
       }
 
       const failedFilenames = errored.map((result) => {
--- a/src/components/VtkThreeView.vue
+++ b/src/components/VtkThreeView.vue
@@ -77,6 +77,7 @@ import { useSceneBuilder } from '../composables/useSceneBuilder';
 import { useViewConfigStore } from '../store/view-configs';
 import { usePersistCameraConfig } from '../composables/usePersistCameraConfig';
 import { Views } from '../config';
+import { useModelStore } from '../store/datasets-models';
 
 export default defineComponent({
   props: {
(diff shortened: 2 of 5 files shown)


865979216:
if i just load one .vtp file，3D view also does not show it

PaulHax:
@865979216

Thanks for trying.  So we can reproduce, can you:
- post a new issue 
- with the image file
- with the .vtp file
- the steps you took?

floryst:
This is an old bit of functionality that was removed because we didn't have any defined capabilities for managing the geometry apart from showing it. This will be re-added back into a future version of VolView.

PR #115, opened 2022-09-09
fix(layouts): Make layouts more generic

tomsuchel:
This fix allows to configure layouts only in the `config.ts` file, without having to modify the `App.vue` template.


floryst:
LGTM!

PR #116, opened 2022-09-12
feat: refactored and improved layout handling

### LINKED ISSUE #111: Selecting layouts resets the camera views
Problem: when selecting a layout, camera settings reset. This is undesired behavior.

Expected: when selecting a layout, the views should visually remain the same.

This issue stems from the view components generating their own IDs. Instead, IDs should be provided by the layout, allowing for view config persistence.

floryst:
Other things of note:
- lps types now moved to the types folder.
- viewID and viewProxy are now refs, so lots of changes are due to that change.
- slice configs got an axisDirection property to track the 2D slice direction.
- simplified proxy definitions

Outstanding issues:
- Synchronization of slice/w&l should happen application-side.
- [x] ~There is a bug with the WidgetManager where layout updates cause an infinite loop.~ Should be resolved in https://github.com/Kitware/vtk-js/pull/2506
- Remove the 3D view store (to be subsumed into the view config store).

FYI @cjh1 @PaulHax. The core layout changes are in LayoutGrid, `config.ts`, and the view store. (The 2D view store has been removed.)

Closes #111


cjh1:
@floryst This is looking good. The only concern I have is if we need to be able to support dynamically creating views. Think creating new render views in ParaView. Currently we have a hard coded set of views in config.ts. For example what if we want to be able create an arbitrary number of 3D views? (Currently we can only support one in a layout ). This seems like a use case that users/developers might want?

floryst:
It certainly is a use-case we want to support, and this change is a stepping stone to such. We just need to create another View3D view with a different ID, and it will be dynamically created if a layout uses it.

src/config.ts
```diff
@@ -1,32 +1,44 @@
...
-    objType: 'View2D',
-    key: ViewKey.SagittalView,
-    viewDirection: 'Posterior',
-    viewUp: 'Superior',
+    objType: 'View',
+    viewType: 'View2D',
+    id: 'Sagittal',
+    props: {
+      viewDirection: 'Posterior',
+      viewUp: 'Superior',
```
cjh1, line 22: So I think these values should be include in our state file. Should we add them to the view store, rather than just the id? Obvious we can just look them up based on the key, but from a reproducible standpoint I think they belong in the state file.

floryst: The current layout should be available from the view store. I think a non-cloned reference to one of the layouts gets saved into the view store.

floryst:
FYI @finetjul @tomsuchel this change affects how layouts are handled and how the views are created, which is relevant to the reslice view.

src/types/layout.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface ViewProps {
+  viewDirection: LPSAxisDir;
+  viewUp: LPSAxisDir;
+}
+
+export type LayoutView = {
+  objType: 'View';
+  viewType: 'View2D' | 'View3D';
+  id: string;
+  props: ViewProps;
```
cjh1, line 19: So to me these ( along with the `viewType` ) are properties of the view and don't really belong in the layout. The `items` in a layout should be `Array<Layout | string>` as in a layout or a viewID. Then in the view store rather than saving just a list of id, we should save these props and `viewType` along with the id. Thoughts?

floryst: The latest commit contains the changes to move the view specs into the store. Let me know how that seems to you.

cjh1: Looks good to me!

cjh1 (approved):

floryst:
I can address the outstanding issues in separate PRs.

PR #124, opened 2022-09-19
refactor: move view3D store to view config store

floryst:
Subsumes the view 3D store into the view config store. 

@cjh1 @paulhax


cjh1 (approved):
LGTM

cjh1:
@floryst There seems to be a problem with this, if you switch views, the 3d view doesn't get restored?

cjh1:
The Axial view also goes blank ...

cjh1:
Works for me :+1:

PR #125, opened 2022-09-21
feat(TwoView): sync windowing parameters

floryst:
Adds window & level syncing back in at the app level, and not the proxy level (which made it more difficult to track the current window/level properties).

### DIFF
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -69,7 +69,12 @@
         <template v-slot:bottom-left>
           <div class="annotation-cell">
             <div>Slice: {{ slice + 1 }}/{{ sliceMax + 1 }}</div>
-            <div>
+            <div
+              v-if="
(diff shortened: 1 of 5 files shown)


src/components/VtkTwoView.vue
```diff
@@ -563,7 +579,7 @@ export default defineComponent({
     // --- apply windowing and slice configs --- //
 
     watchEffect(() => {
-      if (sliceConfig.value === null || wlConfig.value === null) {
+      if (!sliceConfig.value || !wlConfig.value) {
```
cjh1, line 582: Isn't it better to explicit the `null` ness?

floryst: This can be reverted back to explicit null checks. Thanks for catching that.

floryst:
Re-based and ready.

cjh1 (approved):
Looks magically good!

PR #127, opened 2022-09-22
fix(slice): ensure slice configure is preserved

cjh1:


floryst:
LGTM! We can worry about the case where the image data changes later.

PR #128, opened 2022-09-22
feat(state): add view config to state file

### LINKED ISSUE #112: Saving of layout and view configuration
Currently the state saving only saves data. The next step is to save the view layout and configuration.

cjh1:
Closes #112

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -251,7 +251,7 @@ export default defineComponent({
 
     // --- layout --- //
 
-    const layoutName: Ref<keyof typeof Layouts> = ref('QuadView');
+    const layoutName: Ref<string> = ref('Quad View');
     const { layout: currentLayout } = storeToRefs(viewStore);
 
     watch(
@@ -264,6 +264,15 @@ export default defineComponent({
         immediate: true,
       }
     );
+
+    watch(currentLayout, () => {
+      if (
+        currentLayout.value?.name &&
+        currentLayout.value.name !== layoutName.value
+      ) {
+        layoutName.value = currentLayout.value.name;
+      }
+    });
 
     // --- file handling --- //
     async function openFiles(files: FileList | null) {
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -165,6 +165,7 @@ import {
   defineComponent,
   onBeforeMount,
   onBeforeUnmount,
+  onUnmounted,
   onMounted,
   PropType,
   ref,
@@ -203,7 +204,7 @@ import { LPSAxisDir } from '../types/lps';
 import { ViewProxyType } from '../core/proxies';
 import { useViewProxy } from '../composables/useViewProxy';
 import { useWidgetManager } from '../composables/useWidgetManager';
-import { CameraConfig } from '../store/view-configs/camera';
+import { CameraConfig } from '../store/view-configs/types';
 import { defaultSliceConfig } from '../store/view-configs/slicing';
 import { useWindowingSync } from '../composables/sync/useWindowingSync';
 
--- a/src/io/state-file/schema.ts
+++ b/src/io/state-file/schema.ts
@@ -1,5 +1,32 @@
 import JSZip from 'jszip';
 import { z } from 'zod';
+import { Vector3 } from '@kitware/vtk.js/types';
+import {
+  PiecewiseGaussian,
+  PiecewiseNode,
+} from '@kitware/vtk.js/Proxy/Core/PiecewiseFunctionProxy';
+
+import vtkPiecewiseFunctionProxy from '@kitware/vtk.js/Proxy/Core/PiecewiseFunctionProxy';
+
+import {
+  CameraConfig,
+  SliceConfig,
+  WindowLevelConfig,
+  VolumeColorConfig,
+} from '../../store/view-configs/types';
+import { LPSAxisDir } from '../../types/lps';
+import { LayoutDirection } from '../../types/layout';
+import {
+  ViewType,
+  ColorBy,
+  ColorTransferFunction,
+  OpacityFunction,
+  OpacityGaussians,
+  OpacityPoints,
+  OpacityNodes,
+  ColoringConfig,
+  CVRConfig,
+} from '../../types/views';
 
 export enum DataSetType {
   DICOM = 'dicom',
@@ -8,6 +35,15 @@ export enum DataSetType {
 
 const DataSetTypeNative = z.nativeEnum(DataSetType);
 
+const LPSAxisDir: z.ZodType<LPSAxisDir> = z.union([
+  z.literal('Left'),
+  z.literal('Right'),
+  z.literal('Posterior'),
+  z.literal('Anterior'),
+  z.literal('Superior'),
+  z.literal('Inferior'),
+]);
+
 const DataSet = z.object({
   id: z.string(),
   path: z.string(),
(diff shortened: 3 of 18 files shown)


cjh1:
@floryst This is ready to try. The one issue I do see is that the restore 3D view need a interaction to re-render, missing a re-render call somewhere?

floryst:
We may need a render call in the 3D view once the parameters have been restored.

floryst:
> @floryst This is ready to try. The one issue I do see is that the restored 3D view need a interaction to re-render, missing a re-render call somewhere?

I'm actually not seeing this behavior. What are you doing to see this?

src/components/VtkTwoView.vue
```diff
@@ -362,6 +363,10 @@ export default defineComponent({
       viewProxy.value.getInteractorStyle2D().removeAllManipulators();
     });
 
+    onUnmounted(() => {
+      viewConfigStore.removeView(viewID.value);
```
floryst, line 367: I think VtkThreeView needs this as well.

src/io/state-file/schema.ts
```diff
@@ -1,5 +1,38 @@
 import JSZip from 'jszip';
 import { z } from 'zod';
+import { Vector3 } from '@kitware/vtk.js/types';
+import {
+  // IPiecewiseFunctionProxyMode,
```
floryst, line 5: Remove the extra comment.

cjh1:
> > @floryst This is ready to try. The one issue I do see is that the restored 3D view need a interaction to re-render, missing a re-render call somewhere?
> 
> I'm actually not seeing this behavior. What are you doing to see this? I opened two sample datasets, rotated the 3D views, and switched between the two. The results are as expected.

Strange, just pushed a fix for it :smile:

src/io/state-file/schema.ts
```diff
@@ -1,5 +1,38 @@
...
+  CVRConfig,
+} from '../../types/views';
+
+// This is duplicated from shims.vtk.d.ts as we can't use the enum values at
+// runtime as they are stripped.
+export enum IPiecewiseFunctionProxyMode {
+  Gaussians = 0,
+  Points = 1,
+  Nodes = 2,
+}
```
floryst, line 35: I think you want to use `vtkPiecewiseFunctionProxy.Modes` instead of this enum, e.g. `vtkPiecewiseFunctionProxy.Modes.Gaussians`.

src/store/view-configs/common.ts
```diff
@@ -16,3 +23,45 @@ export const removeDataFromConfig =
       config.deleteSecondKey(dataID);
     }
   };
+
+export const ensureDefault = (key: string, records: Record<string, any>) => {
```
floryst, line 27: This could be more generally useful in utils, e.g.

```typescript
export const ensureDefault = <T>(key: string, records: Record<string, T>, default: T) => {
  if (!(key in records)) records[key] = default;
})

ensureDefault(key, config, {})
```

This assumes that the type of config is known, and not assumed to be `any`.

src/io/state-file/schema.ts
cjh1: Ok, good to know

src/store/view-configs/common.ts
cjh1: Makes sense ...

cjh1: Ok, good idea

src/io/state-file/schema.ts
cjh1: Great!

cjh1: This don't work, vtkPiecewiseFunctionProxy still refers to an enum defined in a d.ts file which will be stripped.

floryst: I've used `vtkPiecewiseFunctionProxy.Modes.[Name]` as a value in other places, and it's backed by an actual object that maps those names to values. In what way is it getting stripped?

cjh1: So the issue is as I understand it that even with `vtkPiecewiseFunctionProxy.Modes` its type is defined as an `enum` ( `IPiecewiseFunctionProxyMode` ) inside `shim.vtk.d,ts`. Now unlike other types such as `interface` an enum in TypeScript has complied output ( the JavaScript implementation ). `.d.ts` are not explicitly imported so TypeScript does not have a place to inject the generated code. So if you try to refer to them at runtime for example `console.log(vtkPiecewiseFunctionProxy.Mode.Gaussians)` you will get a runtime error like this:

```
bootstrap:27 Uncaught TypeError: Cannot read properties of undefined (reading 'Mode')
    at ./src/io/state-file/schema.ts (schema.ts:38:13)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./src/store/datasets-images.ts (datasets-files.ts:11:8)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./src/store/datasets-dicom.ts (zip.ts:4:8)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./src/store/datasets.ts (datasets-models.ts:10:8)
```
If its back by a real object, it appears that TypeScript is still just using the type its finds in the `.d.ts` file.

src/io/state-file/schema.ts
cjh1: Removed

src/components/VtkTwoView.vue
cjh1: Added

src/io/state-file/schema.ts
floryst: Oddly enough, I can do the following just fine:

```js
import vtkPiecewiseFunctionProxy from '@kitware/vtk.js/Proxy/Core/PiecewiseFunctionProxy';
console.log(vtkPiecewiseFunctionProxy.Mode.Gaussians);
```

I don't think I have any local changes that can change this behavior, but I can check out a new copy to see.

cjh1: Ok, I see the problem, I was adding to the import inside `{ } from '@kitware/vtk.js/Proxy/Core/PiecewiseFunctionProxy'`, using the default import syntax does indeed resolve the issue.

src/store/views.ts
```diff
@@ -76,5 +82,62 @@ export const useViewStore = defineStore('view', {
...
+          config,
+        };
+
+        views.push(view);
+      });
+
+      // Serialize the view config
+      viewConfigStore.serialize(stateFile);
+    },
+    deserialize(views: View[], dataIDMap: Record<string, string>) {
```
floryst, line 124: I noticed we don't restore the layout here. I think we will also need to store the name of the layout in the store as well, since we need to ensure the layout selector is correct.

cjh1: Oops, yes we are serializing it be never restoring it.

floryst:
We need to also save/restore the name of the layout. When I switch to axial primary, save, and then restore, the layout button is still on "Quad".

cjh1:
> We need to also save/restore the name of the layout. When I switch to axial primary, save, and then restore, the layout button is still on "Quad".

We are saving and restoring it as part of the layout, not sure why its not updating. I can take a look.

floryst:
There seems to be a bug with the Window/Level controls. When I click and drag from right to left, the w/l resets. I can't repro this on master.

cjh1:
@floryst Is this with a restored data set? Or you are loading a "fresh" dataset?

floryst:
Fresh dataset. I loaded up Abdomen-MRI and the Head-MRA dataset at the same time, waited, then tried W/L.

src/components/App.vue
```diff
@@ -265,6 +265,17 @@ export default defineComponent({
       }
     );
 
+    watch(currentLayout, () => {
+      console.log('looper!');
+      console.log(currentLayout.value.name);
```
floryst, line 270: Remove debugging statements.

cjh1: Oops!

cjh1: Done

PR #134, opened 2022-09-29
Save and restore tool state

### LINKED ISSUE #114: Saving of tool state and labelmaps
In addition to saving view configurations, we would want to save tools and labelmaps.

cjh1:
This is branched of #128 which should go in first ...

Closes #114.

### DIFF
--- a/src/io/state-file/index.ts
+++ b/src/io/state-file/index.ts
@@ -3,6 +3,9 @@ import { saveAs } from 'file-saver';
 import { useFileStore } from '@/src/store/datasets-files';
 import { LayoutDirection } from '@/src/types/layout';
 import { useViewStore } from '@/src/store/views';
+import { useLabelmapStore } from '@/src/store/datasets-labelmaps';
+import { useToolStore } from '@/src/store/tools';
+import { Tools } from '@/src/store/tools/types';
 import {
   useDatasetStore,
   makeFileFailureStatus,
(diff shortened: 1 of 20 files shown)


src/io/state-file/index.ts
```diff
@@ -18,19 +21,35 @@ import {
...
+      crosshairs: {
+        position: [0, 0, 0],
+      },
+      paint: {
+        activeLabelmapID: null,
+        brushSize: 8,
+        brushValue: 1,
+        labelmapOpacity: 1,
+      },
+      current: Tools.WindowLevel,
```
floryst, line 51: Does it make sense to save the current tool? I think the data is the most important thing. (This also means we don't need to save the crosshairs state, since it will be overwritten the next time it's activated anyways.) @aylward

floryst: I suppose this doesn't matter for now. Saving this state is mostly inconsequential. We can adjust later as needed.

cjh1: Personally I think saving the current tool is nice, as it more complete restores the state of your previous session, however, I can be persuaded to remove it.

src/store/datasets-labelmaps.ts
```diff
@@ -58,5 +80,58 @@ export const useLabelmapStore = defineStore('labelmap', {
 
       return id;
     },
+    async serialize(state: StateFile) {
+      const { labelMaps } = state.manifest;
+      const { zip } = state;
+
+      // const serializer = vtkXMLImageDataWriter.newInstance();
```
floryst, line 87: Extraneous comment

floryst:
Overall, LGTM!

Issue #135, opened 2022-09-29
DICOM thumbnails sometimes do not load


floryst:
Repro steps:
- Load a local dicom dataset, and load the Abdomen-MRI sample dataset
- expand all studies to show the thumbnails for each volume
- Some thumbnails will not render

The immediate cause is due to `file` being undefined on this line: https://github.com/KitwareMedical/VolView/blob/master/src/io/dicom.ts#L138

cjh1:
I can create this and will take a look

PR #136, opened 2022-09-29
fix(dicom): clean up dicom files

cjh1:
So the assumption that itk-wasm clean ups the files it copies over is wrong! Add the necessary clean up.

Fixes: #135


floryst:
LGTM! Good catch!

PR #142, opened 2022-10-03
Fix volume transfer functions

### LINKED ISSUE #132: CT-AAA color shift doesn't match desired defaults
@aylward

floryst:
This fixes the rendering of the volume transfer functions, mostly for the medical presets. FYI @aylward 

- [x] update UX for color mapping range shifter. Maybe we will apply the shift to both the color transfer function and the opacity points as well.

Fixes #132


floryst:
@cjh1 @PaulHax this is ready, if you'd like to look over it. The refactoring commits just move some code around for maintainability.

PR #147, opened 2022-10-06
feat(VolumeRendering): default colors via modality

### LINKED ISSUE #131: Detect DICOM modality and filter and set default transfer functions
@aylward if you have any details you'd like to add, feel free to.

floryst:
This will select a default preset based on the modality. Currently, only CT and MR are supported.

Closes #131 

FYI @aylward @cjh1 @PaulHax

### DIFF
--- a/src/store/datasets-images.ts
+++ b/src/store/datasets-images.ts
@@ -5,12 +5,10 @@ import { vtkImageData } from '@kitware/vtk.js/Common/DataModel/ImageData';
 
 import { defaultLPSDirections, getLPSDirections } from '../utils/lps';
 import { removeFromArray } from '../utils';
-import { useViewConfigStore } from './view-configs';
 import { StateFile, DataSetType, DataSet } from '../io/state-file/schema';
 import { serializeData } from '../io/state-file/utils';
 import { FILE_READERS } from '../io';
 import { useFileStore } from './datasets-files';
-import { useViewStore } from './views';
 import { LPSDirections } from '../types/lps';
 
 export interface ImageMetadata {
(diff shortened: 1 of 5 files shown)


src/store/view-configs/volume-coloring.ts
```diff
@@ -91,26 +111,41 @@ export const setupVolumeColorConfig = () => {
...
   const updateVolumeCVRParameters = createUpdateFunc('cvr');
 
+  const setVolumeColorPreset = (
+    viewID: string,
+    imageID: string,
+    preset: string
+  ) => {
+    const imageStore = useImageStore();
+    const image = imageStore.dataIndex[imageID];
+    if (!image) return;
```
PaulHax, line 121: `console.warn()` here?  Or comment on case when we expect no image?

src/store/datasets-images.ts
```diff
@@ -86,14 +84,6 @@ export const useImageStore = defineStore('images', {
...
         del(this.metadata, id);
         removeFromArray(this.idList, id);
-
-        // Remove the data views
-        const viewStore = useViewStore();
-        const viewConfigStore = useViewConfigStore();
-
-        viewStore.viewIDs.forEach((viewID: string) => {
-          viewConfigStore.removeData(id, viewID);
-        });
```
PaulHax, line 96: =)

PaulHax (approved):
Good defaults are nice.  Did we want to filter the big list of presets in this PR or save for later?

src/store/datasets-images.ts
floryst: Finally (re)moved! Had to avoid cyclic dependencies.

src/store/view-configs/volume-coloring.ts
floryst: There isn't much of a chance that this would occur. Realistically, throwing would be more appropriate here, but the one valid case is if the dataset somehow gets deleted prior to setVolumeColoPreset (e.g. in a promise).

floryst: > Good defaults are nice. Did we want to filter the big list of presets in this PR or save for later?

Good call on the filtering. Do we want to filter @aylward? I think it would be useful to have all presets available regardless of the modality.

floryst:
Merging for now. We can filter presets in a later PR.

PR #154, opened 2022-10-13
feat(samples): add new sample datasets

floryst:
Adds CT Neck and CT chest examples. @cjh1 @PaulHax

@aylward please let me know if you want to update the descriptions.


PaulHax (approved):
Works good.  

I kept clicking CT-neck trying to load it, and nothing was happening... But it was already loaded 🤦  
Keeping the check-mark on already clicked samples might help me.

PaulHax:
> That's a good point. Loading sample data should then focus whatever you're loading.

Good idea.  Would not have helped past goober me as CT-Neck was "focused" (which leads to another UX tweek: display focused dataset name somewhere.)

floryst:
I've updated the PR to prevent re-downloading a sample dataset when you've already loaded it. Additionally, I've removed the non-oriented datasets.

PaulHax:
> I've updated the PR to prevent re-downloading a sample dataset when you've already loaded it.

The sample dataset images are still clickable and triggering downloads.

floryst:
The downloaded samples are now not clickable.

PaulHax:
Nailed it.  Sample Data experience is slick.

PR #155, opened 2022-10-13
feat(messages): support message persistence

floryst:
This allows for the loading message to persist instead of disappearing after a timeout.

@cjh1 @PaulHax

### DIFF
--- a/src/store/messages.ts
+++ b/src/store/messages.ts
@@ -8,16 +8,20 @@ export enum MessageType {
   Success,
 }
 
+export type MessageOptions = {
+  details?: string;
+  persist: boolean;
+};
+
 export interface Message {
   id: string;
   type: MessageType;
   title: string;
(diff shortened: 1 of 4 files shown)


src/store/messages.ts
```diff
@@ -79,10 +123,22 @@ export const useMessageStore = defineStore('message', {
...
+    ) {
       const id = String(this._nextID++);
+      let options: MessageOptions = { persist: false };
+      if (details) {
+        if (typeof details === 'string') {
+          options.details = details;
+        } else {
+          options = details;
+        }
+      }
```
PaulHax, line 138: Another way:
```
const options: MessageOptions = {
  persist: false,
  ...(typeof details === 'string' ? { details } : details),
};
```

PaulHax (approved):
Verks good!

PR #160, opened 2022-10-17
ENH: Improve render quality by descreasing sample distance

aylward:
The heuristic for sample distance computation is too conservative
and produced significant speckle noise in the renderings.   This
commit decreases sample distance and image quality become quite good.


floryst (approved):
LGTM!

PR #161, opened 2022-10-17
BUG: join does not need a wildcard and specifying one breaks on Windows

aylward:
On Windows, if you try to run the server, you get the error

 unable to locate 'C:\src\VolView\src\io\itk-dicom\web-build\dicom*' glob

 ERROR in unable to locate
 'C:\src\VolView\src\io\itk-dicom\web-build\dicom*' glob

This is used to copy files for the dist - copying files that start with dicom in the web-build dir.  However, every file in that dir starts with dicom, so you can just specify the top level dir, web-build.  Confirmed to work on Windows.


vue.config.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
             },
           },
           {
-            from: path.join(__dirname, 'src', 'io', 'itk-dicom','web-build', 'dicom*'),
+            from: path.join(__dirname, 'src', 'io', 'itk-dicom','web-build'),
```
floryst, line 63: The glob is meant to avoid copying intermediate build artifacts. It's more relevant when building the dicom-handling code. Does the following suggestion work on Windows? It might be that the glob module only handles forward slashes, independently of the platform.

```suggestion
            from: `${__dirname}/src/io/itk-dicom/web-build/dicom*`,
```

vue.config.js
```diff
@@ -60,8 +60,11 @@ module.exports = {
             },
           },
           {
-            from: path.join(__dirname, 'src', 'io', 'itk-dicom','web-build', 'dicom*'),
-            to: path.join(__dirname, 'dist', 'itk', 'pipelines', '[name][ext]')
+            from: path.join(__dirname, 'src', 'io', 'itk-dicom','web-build',),
```
floryst, line 63: Trailing comma in params list

PR #162, opened 2022-10-18
ENH: Reorganized RenderingModule GUI and add camera light

aylward:
Addresses Issues https://github.com/KitwareMedical/VolView/issues/158 and https://github.com/KitwareMedical/VolView/issues/156

The color/opacity transfer function is always shown at the top of the rendering module, and there is now the option to have the light follow the camera (using HeadLight).   Should eventually re-think organization of light controls to simplify them.


floryst:
LGTM overall. Toggling the fixed camera on doesn't cause the 3D view to render until you click into the view.

floryst:
I can fix the rendering hiccup since I will be making fixes to avoid flickering with LAO enabled.

PR #166, opened 2022-10-18
BUG: Remove slider appearing next to transfer func graph

aylward:
Graph's frame now fits in x and y.  Also removed extra (redundant) padding in x.


floryst:
LGTM!

PR #167, opened 2022-10-18
Error handling

floryst:
- Catch errors that bubble up to the window
- improve messages UI
- minor refactoring for types and linting

@cjh1 @PaulHax


PaulHax:
Works good but a while scroll bar appeared:

image

floryst:
> Works good but a wild scroll bar appeared:

The wild scrollbar actually is fixed in #166!

PR #169, opened 2022-10-18
ENH: Add option for light to follow camera and otherwise remain fixed

aylward:
When option is enabled, light is a headlight and moves as camera moves.

When disabled, position of light is fixed if the camera is moved.

Allows the user to specify how shadows should be cast.

### DIFF
--- a/src/store/view-configs/volume-coloring.ts
+++ b/src/store/view-configs/volume-coloring.ts
@@ -52,7 +52,7 @@ export const defaultVolumeColorConfig = (): VolumeColorConfig => ({
   },
   cvr: {
     enabled: false,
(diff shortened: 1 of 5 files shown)


src/store/view-configs/volume-coloring.ts
```diff
@@ -52,8 +52,7 @@ export const defaultVolumeColorConfig = (): VolumeColorConfig => ({
   },
   cvr: {
     enabled: false,
-    flipLightPosition: false,
-    fixedLightPosition: false,
+    lightFollowsCamera: false,
```
floryst, line 55: I would suggest having the light following the camera by default, otherwise the default light location won't make sense.

PR #174, opened 2022-10-25
ENH: Added 3D ultrasound example, of a baby

aylward:
Added data from tomovision.com


floryst:
This is an MHA, so the `volumeKey` doesn't apply here. I will need to figure out a better way of tracking datasets that don't have a volume key.

src/config.ts
```diff
@@ -174,6 +175,16 @@ export const SAMPLE_DATA: SampleDataset[] = [
...
       '1.3.6.1.4.1.14519.5.2.1.7311.5101.206828891270520544417996275680.5tse2d1254.538438420111018.1D000000SN0D000000S0D000000S0D000000S0D970296SN0D241922',
   },
+  {
+    name: '3D US Fetus',
+    filename: '3DUS-Fetus.mha',
+    description:
+      '3D ultrasound of a baby. Downloaded from tomovision.com.(8 MB)',
+    url: 'https://data.kitware.com/api/v1/item/635679c311dab8142820a4f4/download',
+    image: USFetusThumbnail,
+    volumeKey:
```
floryst, line 185: We can remove the volumeKey now.

PR #199, opened 2022-11-01
feat(SampleDataBrowser): detect whether a sample is loaded

floryst:
This improves sample detection by relying on the loaded data ID instead of just the volume key, allowing for sample datasets that are not DICOM.


PaulHax:
Works with the new ultrasound baby: https://github.com/PaulHax/VolView/tree/ultra-baby-on-sample-detection

PR #203, opened 2022-11-03
Disable CVR while animating

### LINKED ISSUE #195: Poor volume rendering performance on large datasets
Poor volume rendering performance on large datasets (CTA chest). This is understandable, but perhaps more tuning could be done for managing quality <--> performance tradeoff.

For e.g. CTA chest 3D view is very sticky. When trying to rotate the view, it takes a second or two for the view to start responding to the click-drag action. Similarly, a related issue is that LAO kernel size/radius sliders are hard to drag. They should only update the rendering upon mouse button release.

floryst:
This disables CVR while the 3D view is animating. I will also investigate disabling CVR while cropping.

@jadh4v please play with this to see if performance is any better. You can use the netlify link at the bottom of the Github Actions checks.

Closes #195.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "hasInstallScript": true,
       "dependencies": {
-        "@kitware/vtk.js": "^25.8.2",
+        "@kitware/vtk.js": "^25.13.0",
         "@vue/composition-api": "1.6.3",
         "@vueuse/core": "^8.5.0",
         "@vueuse/shared": "^9.2.0",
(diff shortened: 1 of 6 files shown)


jadh4v:
The rotate action is very sticky on the chest CT volume. It takes quite some time to kick in, but it works well once the grab action takes effect and the volume starts rotating. Something tells me there's some bottleneck in the way the user interaction starts, which is less noticeable on smaller volumes for some reason.

https://user-images.githubusercontent.com/22624785/199871346-9993b738-a30f-4888-8e9c-f81c403d6ebf.mp4

floryst:
This appears to be the widget manager capture phase that's causing the initial slowdown. I wonder how much CVR is affecting the hardware selector pass. In any case, it should be disabled.

jadh4v:
> This appears to be the widget manager capture phase that's causing the initial slowdown. I wonder how much CVR is affecting the hardware selector pass. In any case, it should be disabled.

Agreed. I think the reduction in volume rendering quality to improve interaction is fairly aggressive. The stickiness could perhaps be originating from other parts.

floryst:
The following WidgetManager fixes should help improve this performance on the vtk.js side: https://github.com/Kitware/vtk-js/pull/2623

PR #212, opened 2022-11-08
feat(VolumeRendering): Enable cinematic rendering by default

### LINKED ISSUE #180: Change the default volume view to cinematic render
How about changing the default view to show enhanced scattered images?

At the very least, change the default value of the `Blending` slider to the middle. Right now, the view doesn't change at all after enabling cinematic rendering.

sankhesh:
Fixes #180


src/store/view-configs/volume-coloring.ts
```diff
@@ -53,8 +53,8 @@ export const defaultVolumeColorConfig = (): VolumeColorConfig => ({
   cvr: {
     enabled: true,
     lightFollowsCamera: true,
-    useVolumetricScatteringBlending: false,
-    volumetricScatteringBlending: 0,
+    useVolumetricScatteringBlending: true,
```
floryst, line 56: LGTM! Do we want to have scattering enabled by default? My only concern is that the initial render will take longer as a result. @aylward

sankhesh: IMHO, scattered rendering is the first thing that a user should see. To improve performance, we should disable interactive shading. As for first render, do you know the time difference between cinematic and shaded volume first frames? If that is of  bigger concern, we could look into improving first frame time separately

PR #229, opened 2022-11-09
fix(ManipulatorTool): retain default mouse action across tools

jadh4v:
Allow the default mouse actions (panning, zooming) to continue working with modifier keys even when the primary interaction tool (bound to left-mouse click-drag) is changed through the toolbar.

Addresses Issue: https://github.com/KitwareMedical/VolView/issues/172

### DIFF
--- a/src/components/tools/ManipulatorTool.vue
+++ b/src/components/tools/ManipulatorTool.vue
@@ -2,6 +2,7 @@
 import {
   computed,
   defineComponent,
+  PropType,
   onBeforeUnmount,
   toRefs,
   watch,
@@ -22,8 +23,8 @@ export default defineComponent({
       required: true,
     },
     options: {
-      type: Object,
-      default: () => ({}),
+      type: Array as PropType<Array<any>>,
+      default: () => [],
     },
   },
   setup(props) {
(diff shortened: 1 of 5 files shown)


floryst:
FYI you will need to run `npm run prettify` to conform to the coding style.

jadh4v:
> FYI you will need to run `npm run prettify` to conform to the coding style.

Ok, updated the branch. For some reason it also touched `src/components/styles/utils.css`, even though I never modified it.

src/components/tools/ManipulatorTool.vue
```diff
@@ -38,38 +39,98 @@ export default defineComponent({
...
+        newManipulators.forEach((m) => {
+          if (m.isA('vtkCompositeMouseManipulator')) {
+            style.addMouseManipulator(m);
+          }
+        });
+      }
+    }
 
+    /*
     watch(options, (newOptions) => {
```
jadh4v, line 68: This commented function is an alternative approach. It will delete/create all mouse manipulators rather than updating existing ones and only inserting/deleting when needed.

src/components/tools/ManipulatorTool.vue
```diff
@@ -38,38 +39,98 @@ export default defineComponent({
         : viewProxy.value.getInteractorStyle3D();
     });
 
-    const manipulator = props.manipulatorClass.newInstance(options.value);
+    const manipulators: any[] = [];
+    options.value.forEach((opt) =>
+      manipulators.push(props.manipulatorClass.newInstance(opt))
+    );
+
```
floryst, line 46: Instead of doing the initialization here, maybe let the `watch(options, ...)` block do it. You can add a 3rd parameter, `{ immediate: true }`, which is effectively the same as an init call.

src/components/tools/ManipulatorTool.vue
```diff
@@ -38,38 +39,98 @@ export default defineComponent({
...
+      while (manipulators.length > newOptions.length) {
+        const m = manipulators.pop();
+        intStyle.value.removeMouseManipulator(m);
+        m.delete();
       }
+
+      const oldManipulators = Array.from(manipulators);
+
+      let i = manipulators.length;
+      while (manipulators.length < newOptions.length) {
```
floryst, line 96: Can be simplified to `for (let i = manipulators.length; i < newOptions.length; i++)`

src/components/tools/ManipulatorTool.vue
floryst: This approach does have a (small) perf overhead, due to the deletion/creation of more objects. While this approach is cleaner, I think we can keep the current impl.

src/components/tools/ManipulatorTool.vue
```diff
@@ -38,38 +39,79 @@ export default defineComponent({
...
-      if ('control' in newOptions) {
-        manipulator.setControl(newOptions.control);
+    function updateStyle(
+      style: any,
+      newManipulators: any[],
+      oldManipulators: any[]
+    ) {
+      if (style !== null) {
+        oldManipulators.forEach((m) => {
+          if (m.isA('vtkCompositeMouseManipulator')) {
```
floryst, line 51: I don't think we need these type checks, since if someone passes in a non-mouse manipulatorClass, the watch block would throw an error when setting the options anyways.

src/components/tools/ManipulatorTool.vue
```diff
@@ -22,8 +23,8 @@ export default defineComponent({
       required: true,
     },
     options: {
-      type: Object,
-      default: () => ({}),
+      type: Array as PropType<Array<any>>,
```
floryst, line 26: We should probably strictly type this, but we can do that later. (We would also look to strictly typing the manipulatorClass.)

floryst:
Once you think this PR is ready, rebase and then we can merge!

floryst:
@jadh4v is this good to go?

jadh4v:
> 

Working on it right now. I'll ping you when its ready.

src/components/tools/ManipulatorTool.vue
jadh4v: ok, auto-created an issue regarding your comment so we can track this in the future.

src/components/tools/ManipulatorTool.vue
jadh4v: Remove the `isA` checks

jadh4v:
@floryst This is ready to go, but please not the additional commit for renaming ManipulatorTool --> MouseManipulatorTool.

jadh4v:
Update: Did a rebase with master.

floryst:
LGTM!

PR #230, opened 2022-11-09
fix(ThreeView): cap the number of samples per ray

floryst:
This limits the max number of samples per ray. The goal is to minimize GPU crashes, as having more than 2000 samples per ray seems to cause crashes. The underlying cause is unclear at this time.


src/components/VtkThreeView.vue
```diff
@@ -323,16 +323,26 @@ export default defineComponent({
...
+          )
+        );
 
-        mapper.setSampleDistance(sampleDistance / 10);
+        let sampleDistance = (1.5 * vec3.length(spacing)) / 10;
+        if (
+          spatialDiagonal / sampleDistance >
+          mapper.getMaximumSamplesPerRay()
+        ) {
+          sampleDistance = spatialDiagonal / mapper.getMaximumSamplesPerRay();
```
aylward, line 343: Perhaps divide by (mapper.getMaximumSamplesPerRay()+1); to avoid precision resulting in 1 too many samples.

aylward: No more crashes for me on Chrome on Windows!

I do see the warning message persists:  macros.js:60 The number of steps required 1501 is larger than the
        specified maximum number of steps 1500.
        Please either change the
        volumeMapper sampleDistance or its maximum number of samples.

aylward (changes requested):

src/components/VtkThreeView.vue
floryst: I changed it to `maxSamplesPerRay-1`, which handles the off-by-one with the number of steps.

floryst:
If crashes are still happening, I can bump the max samples per ray down to 1250 or 1000 (the default).

aylward (approved):

PR #234, opened 2022-11-10
ENH: Add additional space between data-loading options and secure note

aylward:
Added space so that loading options are confused with note about app being secure.

Also clarified the message about the app being secure.

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -125,11 +125,11 @@
                         <v-icon size="64">mdi-arrow-down-bold</v-icon>
                       </div>
                       <div>Drag &amp; drop your DICOM files.</div>
+                      <br /> <br /> <br />
                       <div class="mt-16">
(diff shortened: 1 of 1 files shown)


src/components/App.vue
```diff
@@ -125,11 +125,13 @@
                         <v-icon size="64">mdi-arrow-down-bold</v-icon>
                       </div>
                       <div>Drag &amp; drop your DICOM files.</div>
+                      <br />
+                      <br />
+                      <br />
```
floryst, line 130: I would suggest changing the below `class="mt-16"` to `class="offset-margin"`. At the bottom of this file inside the `<style scoped>` tag, add the following lines:

```css
.offset-margin {
  margin-top: 128px;
}
```

aylward: Done.

aylward:
Will submit another pull request without master merge.

PR #237, opened 2022-11-11
feat(VtkTwoView): change slice with arrow keys

PaulHax:
Keys active on mouse hover or focus.

### DIFF
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -182,6 +182,7 @@ import {
   watchEffect,
 } from '@vue/composition-api';
 import { vec3 } from 'gl-matrix';
+import { onKeyStroke } from '@vueuse/core';
 
 import { useResizeToFit } from '@src/composables/useResizeToFit';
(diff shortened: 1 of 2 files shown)


aylward (approved):
Nice!

I like being able to use up/down or right/left for scrolling.   Works 1 slice at time, as expected.

src/components/VtkTwoView.vue
```diff
@@ -400,6 +407,48 @@ export default defineComponent({
...
+    };
+
+    const hover = ref(false);
+    const onDocumentKeyDown = (event: KeyboardEvent) => {
+      if (hover.value) onKeyDown(event);
+    };
+    document.addEventListener('keydown', onDocumentKeyDown);
+    onBeforeUnmount(() => {
+      document.removeEventListener('keydown', onDocumentKeyDown);
+    });
```
floryst, line 450: I would recommend using [onKeyStroke](https://vueuse.org/core/onKeyStroke/). We have vue-use as a dependency, which can help cut down on having to define cleanup ourselves.

PaulHax: Excellent, done.

PR #243, opened 2022-11-14
ENH: Add funding information to About box.

aylward:
This work was funded, in part, by the NIH via NIBIB and NIGMS R01EB021396,
NIBIB R01EB014955, NCI R01CA220681, and NINDS R42NS086295


floryst:
Need to run `npm run prettify`

PR #252, opened 2022-11-18
Paint: proper paintbrush scaling

floryst:
Instead of trying to resample the brush stencil, rasterize an ellipse on irregularly spaced images.


aylward (approved):
Awesome!

PR #266, opened 2023-01-23
Load remote urls

floryst:
This adds the ability to load remote datasets via URL parameters.

- [x] Add documentation

### DIFF
--- a/src/composables/useAppLoadingNotifications.ts
+++ b/src/composables/useAppLoadingNotifications.ts
@@ -1,0 +1,73 @@
+import { TYPE } from 'vue-toastification';
+import { useToast } from '@/src/composables/useToast';
+import { ToastID } from 'vue-toastification/dist/types/src/types';
+import { useMessageStore } from '../store/messages';
+
+export function useAppLoadingNotifications() {
+  const toast = useToast();
+  const messageStore = useMessageStore();
+
+  let loadingCount = 0;
+  let toastID: ToastID | null = null;
+  let error: Error | null = null;
+
+  const resetState = () => {
+    loadingCount = 0;
+    toastID = null;
+    error = null;
+  };
+
+  const _setError = (err: Error) => {
+    error = err;
+  };
+
+  const startLoading = () => {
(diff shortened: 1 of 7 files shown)


src/composables/useAppLoadingNotifications.ts
```diff
@@ -0,0 +1,73 @@
...
+        closeOnClick: false,
+      });
+    }
+  };
+
+  const stopLoading = () => {
+    loadingCount--;
+    if (loadingCount === 0 && toastID !== null) {
+      if (error) {
+        toast.dismiss(toastID);
```
PaulHax, line 39: Just for my learning: Why did you not throw a bang in `toast.dismiss(toastID!)` and take the runtime error? If `toastID === null` then startLoading would not have been called, which would be a silent bug now?

src/composables/useAppLoadingNotifications.ts
```diff
@@ -0,0 +1,73 @@
+import { TYPE } from 'vue-toastification';
+import { useToast } from '@/src/composables/useToast';
+import { ToastID } from 'vue-toastification/dist/types/src/types';
+import { useMessageStore } from '../store/messages';
+
+export function useAppLoadingNotifications() {
```
PaulHax, line 6: Nice abstraction!

PaulHax (approved):
Works good!

src/utils/index.ts
```diff
@@ -41,6 +41,19 @@ export async function fetchFileWithProgress(
   return new File([bytes], name);
 }
 
+/**
+ * Fetches a file and returns a Blob.
```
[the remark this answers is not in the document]
floryst: Whoops, that's supposed to be "returns a File".

src/composables/useAppLoadingNotifications.ts
floryst: Just defensive programming. Also, I have another project where the linter complains about non-null assertions, so I likely implicitly followed that here.

PR #269, opened 2023-02-03
Environment variable for publicPath for hosting off URL origin

PaulHax:
Adds `VUE_APP_PUBLIC_PATH` environment variable to set Vue CLI `publicPath` and use as base for `itkConfig` paths.  

If VolView is served at `example.com/static/volview`  can now build VolView with `VUE_APP_PUBLIC_PATH=/static/volview npm run build`

Adds `npm run build:watch` script.

Changes package.json to be npm "packagable".

Bumps `itk-wasm` version.


floryst (approved):
LGTM!

PR #272, opened 2023-02-07
Python server demo

floryst:
This adds a functional Python server to VolView through the "Remote Functions" tab.

Future demonstrations would include remote-loading of datasets and measurements.

- [x] add documentation on how to use and extend the server
- [x] support calling server APIs
- [x] support accessing client stores
- [x] UI for configuring connected server
- [x] demos for synchronous and asynchronous server RPC endpoints
- [x] server supports running standalone or as part of an ASGI framework
- [x] added application debug messaging
- [ ] create button-click installer for python server (automated or leads to webpage for download/docs)


src/core/remote/codecs.ts
```diff
@@ -0,0 +1,59 @@
...
+          object.buffer,
+          object.byteOffset,
+          object.byteLength
+        );
+      }
+      return null;
+    },
+    decode: (data: Uint8Array) => {
+      // force a copy to handle alignment issues.
+      // TODO try to align typed array data when encoding
```
PaulHax, line 40:
```suggestion
```
Tracked #275

PaulHax (approved):
Amazing, succinct, magic!

src/components/ServerModule.vue
```diff
@@ -0,0 +1,172 @@
...
+      medianFilterLoading,
+      hasCurrentImage,
+      medianFilterRadius,
+    };
+  },
+});
+</script>
+
+<template>
+  <div class="overflow-y-auto ma-2 fill-height">
```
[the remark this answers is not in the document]
floryst: I'm opting to just use `overflow-x-hidden` for now, since it seems like an easier fix.

src/components/ModulePanel.vue
```diff
@@ -60,11 +61,11 @@ export const Modules = [
     icon: 'cube',
     component: RenderingModule,
   },
-  // {
-  //   name: 'AI',
-  //   icon: 'robot-outline',
-  //   component: null,
-  // },
+  {
+    name: 'Remote Functions',
```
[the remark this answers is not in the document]
floryst: That's a great suggestion!

src/components/ServerModule.vue
```diff
@@ -0,0 +1,172 @@
...
+          <v-btn
+            @click="doMedianFilter"
+            :loading="medianFilterLoading"
+            :disabled="!ready || medianFilterLoading || !hasCurrentImage"
+          >
+            Run Median Filter
+          </v-btn>
+          <span v-if="!hasCurrentImage" class="ml-4 body-2">
+            No image loaded
+          </span>
```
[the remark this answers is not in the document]
floryst: I'm not wrapping in v-col just yet.

floryst:
@aylward I've updated the demo with a streaming progress example.

src/core/remote/codecs.ts
floryst: I may just remove this TODO and leave it as a note. This is a harder problem due to msgpack upstream not providing the necessary context for how much alignment padding to inject.

floryst:
The python server has been updated and improved. Check out the `user_api.py` file for how to use the new API. Details of the new changes:
- switched from wslink to socket.io for RPC. After chatting with @jourdain, we decided against extending wslink outside its current feature scope.
- added the ability to access client stores from the server. This includes accessing state and calling actions. The median filter example demonstrates this usage.
- Renamed the `rpc` decorator to `expose`. Additionally, the `stream` decorator has been merged into `expose`.
- Both the client and server now support custom serializers and deserializers. Previously, only the server supported them.

server/pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.poetry]
+name = "volview_server"
+version = "0.1.0"
+description = "The VolView Python Server"
+authors = ["Forrest <forrest.li@kitware.com>"]
+license = "Apache 2.0"
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
aylward, line 9: Could change to "^3.8.1"

server/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+itk = "^5.3.0"
+black = "^23.1.0"
+flake8 = "^6.0.0"
+pytest = "^7.2.1"
+numpy = "^1.24.1"
+msgpack = "^1.0.4"
+aiohttp = "^3.8.3"
+python-socketio = "^5.8.0"
+
```
aylward, line 18: Missing dependency:
charset-normalizer = "^3.0.1"

floryst:
I've lowered the python requirement to 3.8.1 and updated the dependencies list.

phtran-dev:
Hi @floryst ,
Would you mind sharing your reason of switching from wslink to socket.io ? Does wslink not suffice the requirements of Volview Server ? In addition, I saw many Framework like : FastAPI or Sanic have Socket feature. Is this possible to integrate them into Volview ?
Thanks

floryst:
Hi @phtran-dev,

One limitation of wslink is that bi-directional RPC is not currently supported, which is necessary to support accessing the client state store from the server. While I could technically add a bi-directional RPC mechanism on top of wslink, it's another layer of complexity and lacks implementation symmetry when compared to using socket.io.

Another consideration is that python-socketio is ASGI compatible, so you can integrate this python server into your ASGI-compatible framework (e.g. sanic or fastapi) or server (e.g. uvicorn). I still need to change the docs to using the volview server as an ASGI middleware. I suppose wslink can be made ASGI compatible (if it isn't already), so this is less of a concern.

phtran-dev:
Hi @floryst ,
Thanks a lot for your quick response. That would be great ideas from your team while making Volview be compatible with frameworks supporting ASGI (like FastApi or Sanic). I have another question. My  team is implement 3D Server which employs wslink as transportation and vtk and backend service. The client side is also using vtk.js which has wslink. The result of project is quite good. However, if I switch to use your Volview framework (where I can reuse your socket transportation), does it need to change the client code ? I am asking because I do not think the vtk.js does support the socket.io . Thanks in advance.

jourdain:
Another point that is important to keep in mind is the fact that wslink is really meant to be used with a dedicated server process per client. And for his reference implementation, @floryst wanted to have a simple server (1 instance that serves many clients). But in general, it might be easier to abstract the VolView client side so a wslink version of that client (with a dedicated server process) could be swapped and used. 

@floryst do you think you can make sure the client side could be swapped? That way @phtran-dev can still rely on his existing infrastructure by just adding a protocol on his server side.

floryst:
I'm not sure about a swappable client yet. My suggestion would be to use an ASGI-compatible framework and register the wslink and volview servers under different routes (e.g. `/wslink` routes to the wslink server and `/volview` routes to the volview server). You'll also probably need to do path stripping before passing requests onto wslink or volview.

phtran-dev:
Hi @floryst ,
I am interested in your architecture when one service an serve multiple socket connections (which is contradict to wslink where one server serves only one socket). However the existing VTK.js (client side) is heavily relying on the wslink to communicate with the server as in this example [https://kitware.github.io/vtk-js/examples/RemoteView.html#Source] . So do we need to modify the source code of VTK.js to be compatible with your new Volview server ? Do you have any example of rendering 3D image using your Volview ?

aylward:
The VUE_APP_REMOTE_SERVER_URL variable in the .env file should also be available in the setting dialog.   If this variable must be defined when the app is launched, then changing the variable in the dialog should cause another pop-up that gives the user a button to reload/restart volview with the new setting applied.

If a .env file doesn't exist, it should be automatically created - when running from github source. 

Instead of displaying tabs (e.g., DICOM WEB and Remote Functions) based on a variable being set, we should test the variable and if they work, display the tabs, otherwise don't.   Could even put "test" buttons next to those variables in the settings dialog.

floryst:
@aylward I force-pushed this branch and updated it to match `main`.
- `VUE_APP_REMOTE_SERVER_URL` is now `VITE_REMOTE_SERVER_URL`, so you will need to update your `.env` accordingly
- Documentation has been moved to the `documentation/` directory. @phtran-dev this includes some deployment strategies that integrate with existing servers.

@phtran-dev VolView currently doesn't support remote rendering. The server is just for remote data processing and control, so it won't conflict at all with your current remote rendering setup.

PaulHax:
I can only get the median filter to work on the "MRI HEART ... Ax FIESTA non gated" sample data.  Running the filer on the other sample images hangs the Python server on the `img = await store.dataIndex[base_image_id]` line. 

https://github.com/Kitware/VolView/pull/272/files#diff-c2c4c80c7d8530042fff10f2bee85bb288a3bd85a2f160ef7f54e8ad07ab5d01R70

jadh4v:
> I can only get the median filter to work on the "MRI HEART ... Ax FIESTA non gated" sample data. Running the filter on the other sample images hangs the Python server on the `img = await store.dataIndex[base_image_id]` line.
> 
> https://github.com/Kitware/VolView/pull/272/files#diff-c2c4c80c7d8530042fff10f2bee85bb288a3bd85a2f160ef7f54e8ad07ab5d01R70
> 
> Chrome on Ubuntu

Same for me on Windows 11 with Edge browser. Median filter works on the image Paul mentioned above but not for any other images.

documentation/content/doc/server.md
```diff
@@ -0,0 +1,283 @@
+title: VolView Server Guide
```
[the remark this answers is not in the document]
floryst: IIRC the docs use that special key to render the document title, so we'll keep it that way.

floryst:
I've updated this branch to resolve the median filter issues. Please take a look again!

PaulHax:
Median filter works on all the sample data now ⭐️

aylward:
I'm getting an error when I issue the npm run build command.  Starting from a fresh checkout of VolView.  npm install works with only warnings.  Build returns these errors (the underline belongs under "length" and "cause" in the following report):

(venv) C:\src\Proj_VolView\VolView>npm run build

> volview@4.1.0 build
> vue-tsc --noEmit && vite build

src/core/remote/chunkedParser.ts:125:28 - error TS2339: Property 'length' does not exist on type 'File | Blob'.
  Property 'length' does not exist on type 'File'.

125     while (offset < binary.length) {
                               ~~~~~~

src/core/remote/chunkedParser.ts:126:61 - error TS2339: Property 'length' does not exist on type 'File | Blob'.
  Property 'length' does not exist on type 'File'.

126       const chunkEnd = Math.min(offset + CHUNK_SIZE, binary.length);
                                                                ~~~~~~

src/core/remote/chunkedParser.ts:180:58 - error TS2322: Type 'unknown' is not assignable to type 'Error | undefined'.

180       throw new Error('Failed to parse chunking info', { cause: err });
                                                             ~~~~~


Found 3 errors in the same file, starting at: src/core/remote/chunkedParser.ts:125

floryst:
I've addressed the build type errors.

aylward:
VolView now compiles, and the addition, trivia, and progress tasks work.   The median filtering still doesn't work - a blue pop-up says "Not connected to server."

This was with the smallest data - MRI prostate.   Will try others.

aylward:
Confirmed does not work with other data.

Interestingly, it seems to be running (consuming CPU), but it never returns.

floryst:
Updates:
- I've simplified the API further and re-organized the examples. To run the examples: `python -m volview_server ./examples/example_api.py`
- Updated the median filter example to run in a subprocess to avoid blocking. @aylward this will resolve your timeout issues
- Improved the FastAPI integration example
- Updated the documentation to reflect the new API

aylward:
We need to break out the documentation into two files: one for developers (this file) and one for the VolView application - for which the installation is done by a button press in the VolView application.   That button press should download and install the server on the local machine (ideally automatically, but perhaps leading to a webpage from which the server can be downloaded and installation instructions are given).

We need to remove mention of "example" in the documentation and code base as well.  This is a fully functioning part of VolView.

I'm going to add creation of installer and such into the checklist as well.

server/volview_server/transformers/image_data.py
```diff
@@ -0,0 +1,117 @@
...
+        raise ConvertError("Cannot convert provided vtk_image to an ITK image") from exc
+
+
+def itk_to_vtk_image(itk_image):
+    """Converts an ITK image to a serialized vtkImageData for vtk.js."""
+    if not type(itk_image).__name__.startswith("itkImage"):
+        raise ConvertError("Provided data is not an ITK image")
+
+    size = list(itk_image.GetLargestPossibleRegion().GetSize())
+    values = itk.GetArrayFromImage(itk_image).flatten(order="C")
```
aylward, line 65: Should this be "GetArrayViewFromImage()" so that the data is shared, rather than copied?

server/methods/methods_api.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class CurrentClientState:
+    def __init__(self):
+        self.common = CommonClientState()
+        self.median_filter = MedianFilterClientState()
+
+
+## The median filter is allocated to a process ##
+async def run_median_filter_process(process_pool, state):
```
aylward, line 49: The benefit of having this in a separate function isn't clear to me.  Seems like these few lines could be moved into the caller.  Admittedly the caller calls this function using an await, but I don't think that nesting of awaits are needed.

floryst: There's no reason to have this in a separate function. I wrote the usage before the implementation, and I never did a clean-up pass.

server/volview_server/transformers/image_data.py
floryst: Yep, good point. Resolved.

floryst:
I aim to merge this in without exposing the Remote tab for now. This will get the core components in, and we can add the Remote tab later once we have everything else in place for it.

jadh4v:
> I aim to merge this in without exposing the Remote tab for now. This will get the core components in, and we can add the Remote tab later once we have everything else in place for it.

Please do check-in the updated remote tab component though, so we can understand the usage for downstream LungAir.

floryst:
> Please do check-in the updated remote tab component though, so we can understand the usage for downstream LungAir.

Yep, the `ServerModule.vue` file is being committed, just not used.

PR #281, opened 2023-02-10
Fixes a stack memory leak and extraneous readDICOMTag XHRs

floryst:
- Testing minor optimizations to a hot loop in vtkOpenGLTexture by reducing the number of object lookups.
- Preloads the `read-dicom-tags` pipeline to avoid multiple fetch calls for the same pipeline when reading in a loop. This is less of a concern for image file readers, but it's still good to note it as a potential issue.
- Restores the stack pointer after every `main()` call, since we do not re-initialize the webworker/emscripten program on every call. Cleans up a stack memory leak. (This makes no assumption on the state of the heap in between main() runs.)

FYI @thewtex @PaulHax the last point above may be a point of discussion for re-using webworkers. Do we continue to treat main() as the entrypoint, or do we instead expose some separate run() function?

### DIFF
--- a/patches/@kitware+vtk.js+26.5.4.patch
+++ b/patches/@kitware+vtk.js+26.5.4.patch
@@ -1,0 +1,25 @@
+diff --git a/node_modules/@kitware/vtk.js/Rendering/OpenGL/Texture.js b/node_modules/@kitware/vtk.js/Rendering/OpenGL/Texture.js
+index e2e24b5..35efba5 100644
+--- a/node_modules/@kitware/vtk.js/Rendering/OpenGL/Texture.js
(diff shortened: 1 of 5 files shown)


PaulHax:
Very interesting learning here!  Regarding `main` vs `run`, I don't know enough about emscripten for an informed opinion.

thewtex:
@floryst great find and ideas! A few thoughts building on your suggestions on how to move forward here: https://github.com/InsightSoftwareConsortium/itk-wasm/issues/760 -- additional thoughts appreciated! Its seems that we could add an option for execution that provides `run` capabilities. My sense is that this would we mesh nicely with usage/ implementation, but we will have to see how the implementation pans out.

PR #288, opened 2023-03-14
Remote state

PaulHax:
### New Features
- Loads a `session.volview.zip` state file if in `files` URL parameter.
- Fetch remotely loaded datasets from URLs in state file, rather than zipping in all files.

### How
- Introduce `DatasetFile` type with copyleft `{ url, remoteFilename }` properties.  
- Files downloaded from url parameters, JSON manifest files, and sample datasets get wrapped as a `DatasetFile`.  
- `datasets-dicom` and `datasets-images` eventualy pass `DatasetFile`s to `datasets-files`.
 - On Save State: DatasetFiles with `url` and `remoteFilename` are saved as "Remote Datasets" in the JSON manifest and their raw data is not zipped into the state file.  `state-file/utils.ts` does this and is effectively has `datasets-files`' serialize and deserialize functions.

### Test Plan

For each scenario, verify:
- Save session has no errors
- Remote datasets are not included in .zip
- Load session zip file has no errors and state is restored

Scenarios
- [ ] Multiple uncompressed remote DICOM files
- [ ] Multiple zipped remote DICOM files

Also:
- [ ] A session zip file is loaded when pointed to by URL files parameter
- [ ] Load files from JSON manifest, save session, verify files are not in zip, verify session zip file loads.

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -296,28 +296,28 @@ async function loadRemoteFilesFromURLParams(
   setError: (err: Error) => void
 ) {
   const urls = wrapInArray(params.urls);
-  const names = wrapInArray(params.names ?? []);
+  const names = wrapInArray(params.names);
+  const fetchParams = urls.map((url, idx) => ({
+    url,
+    remoteFilename:
+      names[idx] ||
+      new URL(url, window.location.href).pathname.split('/').at(-1) ||
+      '',
+  }));
 
   const fetchResults = await Promise.allSettled(
-    urls.map(async (url, index) => {
-      const { pathname } = new URL(url, window.location.href);
-      const name = names[index] || pathname.split('/').at(-1) || '';
-      return fetchFile(url, name);
-    })
+    fetchParams.map(({ url, remoteFilename }) => fetchFile(url, remoteFilename))
   );
 
-  const withUrls = fetchResults.map((result, idx) => ({
+  const withParams = fetchResults.map((result, idx) => ({
     result,
-    url: urls[idx],
+    ...fetchParams[idx],
   }));
 
   const [fulfilled, rejected] = partition(
-    ({ result }) => isFulfilled(result),
-    withUrls
-  ) as [
-    Array<{ url: string; result: PromiseFulfilledResult<File> }>,
-    Array<{ url: string; result: PromiseRejectedResult }>
-  ];
+    ({ result }) => isFulfilled(result) && result.value.size !== 0,
+    withParams
+  );
 
   if (rejected.length) {
     setError(
@@ -327,11 +327,12 @@ async function loadRemoteFilesFromURLParams(
     );
   }
 
-  const datasetFiles = fulfilled.map(({ result: { value: file }, url }) =>
-    makeRemote(url)(file)
+  const datasetFiles = fulfilled.map(({ result, url }) =>
+    makeRemote(url, (result as PromiseFulfilledResult<File>).value)
   );
 
-  loadFiles(datasetFiles, setError);
+  // must await for setError to work
+  await loadFiles(datasetFiles, setError);
 }
 
 export default defineComponent({
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -235,8 +235,12 @@ import { useWebGLWatchdog } from '../composables/useWebGLWatchdog';
 import { useAppLoadingNotifications } from '../composables/useAppLoadingNotifications';
 import { wrapInArray } from '../utils';
 import { fetchFile } from '../utils/fetch';
+import { retypeFile } from '../io';
 
-async function loadFiles(files: FileList, setError: (err: Error) => void) {
+async function loadFiles(
+  files: FileList | File[],
+  setError: (err: Error) => void
+) {
   const dataStore = useDatasetStore();
 
   const loadFirstDataset = !dataStore.primarySelection;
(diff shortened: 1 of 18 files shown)


floryst:
+1 to the File wrapper idea, which can have provenance metadata. Serialization will then store the provenance info instead of the file contents. When loading a session, we would need to wait for remote files to load before finishing session load.

src/components/App.vue
```diff
@@ -295,54 +295,44 @@ async function loadRemoteFilesFromURLParams(
...
-        if (errored) {
-          // force the promise to error out
-          throw new Error();
-        }
-        // only select the first dataset
-        return loadResults[0];
-      }
-      throw fileResult.reason;
-    })
+  const names = wrapInArray(params.names);
```
floryst, line 299: We still need `params.names ?? []` here, since if it's undefined then `names` becomes `[undefined]`.

src/store/datasets-files.ts
```diff
@@ -1,8 +1,55 @@
 import { del, set } from '@vue/composition-api';
 import { defineStore } from 'pinia';
+import { pluck } from '../utils';
+
+export type DatasetPath = string & { __type: 'DatasetPath' };
+export type DatasetUrl = string & { __type: 'DatasetUrl' };
+export type LocalDatasetFile = { file: File };
+export type ZipDatasetFile = LocalDatasetFile & { path: string };
```
floryst, line 8: I think we should pick a different name other than "path". I think it's a bit vague. Maybe `archivePath`? We will need to rewrite its usage (I think it's used in serialization/deserialization).

src/store/datasets-files.ts
```diff
@@ -12,23 +59,23 @@ export const useFileStore = defineStore('files', {
...
-        if (dataID in state.byDataID) {
-          return state.byDataID[dataID];
-        }
-        return null;
-      };
-    },
+    getDatasetFiles: (state) => (dataID: string) =>
+      state.byDataID[dataID] ?? [],
+
+    getFiles: (state) => (dataID: string) =>
```
floryst, line 67: Adding function comments here will help readers differentiate between what `getFiles` and `getDatasetFiles` does. Alternatively, maybe we get rid of `getFiles`, and rewrite usages of that function to use `getDatasetFiles`.

src/io/state-file/index.ts
```diff
@@ -100,18 +101,19 @@ async function restore(state: FileEntry[]): Promise<LoadResult[]> {
...
+    const files = await deserializeDataset(dataset);
 
+    // eslint-disable-next-line no-await-in-loop
     const status = await datasetStore
-      .deserialize(dataSet, files)
+      .deserialize(dataset, files)
       .then((result) => {
         if (result.loaded) {
-          stateIDToStoreID[dataSet.id] = result.dataID;
-          fileStore.add(result.dataID, files);
```
floryst, line 114: Do we not need to add to the file store anymore?

src/io/state-file/index.ts
```diff
@@ -15,13 +14,15 @@ import {
...
   ARCHIVE_FILE_TYPES,
   extractArchivesRecursively,
   retypeFile,
 } from '../io';
 import { FileEntry } from '../types';
 import { Manifest, ManifestSchema } from './schema';
+import { deserializeDatasetFiles } from './utils';
 
 const MANIFEST = 'manifest.json';
 const VERSION = '0.0.1';
```
floryst, line 28: The changes to the schema require a version bump. Let's go to 0.0.2.

src/io/state-file/index.ts
```diff
@@ -35,7 +36,8 @@ export async function save(fileName: string) {
   const zip = new JSZip();
   const manifest: Manifest = {
     version: VERSION,
-    dataSets: [],
+    datasets: [],
+    remoteDatasetFileEntries: {},
```
floryst, line 40: I think the suffix `Entries` is redundant. We can leave it as `remoteDatasetFiles`.

floryst (changes requested):

src/store/datasets-files.ts
PaulHax: Added some basic comments.  After "ingestion," App does not really care about `DatasetFiles` so might be nice to keep the `getFiles` helper around.

src/io/state-file/index.ts
PaulHax: `dicomStore` was adding files to `fileStore` in `importFiles`, `imageStore` was not (relying on wrapping code).  Now `imageStore` adds processed images to `fileStore`. So normalized to have final "datastores" be responsible for adding to `fileStore`

src/io/state-file/index.ts
PaulHax: Agreed.  Cut to `remoteFiles` to avoid overlap with `RemoteDatasetFile` in `datasets-files.ts`

src/io/state-file/schema.ts
```diff
@@ -51,12 +51,20 @@ const LPSAxisDir: z.ZodType<LPSAxisDir> = z.union([
...
   id: z.string(),
   path: z.string(),
-  type: DataSetTypeNative,
+  type: DatasetTypeNative,
 });
-export type DataSet = z.infer<typeof DataSet>;
+export type Dataset = z.infer<typeof Dataset>;
+
+const RemoteFile = z.object({
+  path: z.string(),
```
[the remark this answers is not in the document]
PaulHax: Yes, made it optional and renamed to `archivePath`

floryst (approved):
LGTM!

PR #293, opened 2023-03-27
fix(async): use built-in URL api

PaulHax:
Avoids this error when saving labelmaps:

`Refused to execute script from 'http://localhost:8081/5ccc414e5c4232d9.ts' because its MIME type ('video/mp2t') is not executable.`

### DIFF
--- a/src/io/vtk/async.ts
+++ b/src/io/vtk/async.ts
@@ -1,5 +1,4 @@
 import vtk from '@kitware/vtk.js/vtk';
-import { URL } from 'whatwg-url';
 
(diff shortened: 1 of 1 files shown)


floryst:
Why is there a request to a typescript file at runtime? That shouldn't be happening.

We are using the WHATWG-compliant URL implementation because Blink incorrectly parses `gs://` URLs, so let's see if there's a different solution to resolve this issue before looking at using the built-in URL class.

floryst:
Actually, this is indeed the correct fix, since this is restricted to the async worker. Webpack should actually parse that URL, so we won't run into any browser-specific quirks.

PR #295, opened 2023-03-29
fix(dicom): use file indexes as unique file names

### LINKED ISSUE #294: Loading DICOMs with same file names skips some files
File structure of this zip:
[folders-dicoms-same-names.zip](https://github.com/Kitware/VolView/files/11094517/folders-dicoms-same-names.zip)

```
PT PET WB/
  0.dcm
  1.dcm
CT/
  0.dcm
  1.dcm
  2.dcm
  3.dcm
```

If you drag and drop the zip on VolView, it only loads the CT dataset.  Skips the PT files.

Problem is in `categorizeFiles` step.

PaulHax:
closes #294


src/io/dicom.ts
```diff
@@ -81,12 +84,12 @@ export class DICOMIO {
...
 
     const inputs = await Promise.all(
-      files.map(async (file) => {
+      files.map(async (file, index) => {
         const buffer = await file.arrayBuffer();
         return {
           type: InterfaceTypes.BinaryFile,
           data: {
-            path: file.name,
+            path: index.toString(10), // make each file name unique
```
floryst, line 92: Seems unnecessary to specify the radix when the default is decimal anyways.

src/io/dicom.ts
```diff
@@ -106,7 +109,20 @@ export class DICOMIO {
...
+    // File names are indexes into input files array
+    const volumeToFileIndexes = JSON.parse(
+      (result.outputs[0].data as TextStream).data
+    ) as VolumesToFileNamesMap;
+
+    const volumeToFiles = Object.fromEntries(
+      Object.entries(volumeToFileIndexes).map(([volumeKey, fileIndexes]) => [
+        volumeKey,
+        // unique file indexes to Files
+        fileIndexes.map((fileIndex) => files[parseInt(fileIndex, 10)]),
```
floryst, line 121: Same here. `parseInt` defaults to base 10, so no need to specify it here.

floryst (commented):
LGTM overall. Just a minor nit.

src/io/dicom.ts
PaulHax: Removed the radix.

src/io/dicom.ts
PaulHax: Kept radix here due to eslint rule

floryst: Huh, so there is an eslint rule for that. I'm fine with keeping it then.

floryst:
LGTM.

PR #297, opened 2023-04-07
feat(App): save url parameter POSTs session.zip

PaulHax:
If there is a `save` URL parameter and VolView is build with `VUE_APP_ENABLE_REMOTE_SAVE=true`, VolView's "save" button makes a POST request to that URL with the session.zip file.

Example `save` parameter:
`?save=http://localhost:8080/api/v1/item/6425c4b4994364f4d5dcb7b2/volview`

Example build command:
`VUE_APP_ENABLE_REMOTE_SAVE=true npm run build`

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -50,10 +50,17 @@
                 @click="userPromptFiles"
               />
               <tool-button
+                v-if="!saveHappening"
                 size="40"
                 icon="mdi-content-save-all"
                 name="Save session"
                 @click="handleSave"
+              />
+              <v-progress-circular
+                v-if="saveHappening"
+                indeterminate
+                size="40"
+                color="grey lighten-5"
               />
               <div class="my-1 tool-separator" />
               <v-menu offset-x>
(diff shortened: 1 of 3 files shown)


src/io/state-file/index.ts
```diff
@@ -75,8 +74,9 @@ export async function save(fileName: string) {
   await layersStore.serialize(stateFile);
 
   zip.file(MANIFEST, JSON.stringify(manifest));
-  const content = await zip.generateAsync({ type: 'blob' });
-  saveAs(content, fileName);
+  const blob = await zip.generateAsync({ type: 'blob' });
+
+  return blob;
```
floryst, line 79: Can consolidate this to a return.

src/components/App.vue
```diff
@@ -453,11 +453,32 @@ export default defineComponent({
...
+      if (saveUrl) {
+        const blob = await serialize();
+        fetch(saveUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/zip',
+            'Content-Length': blob.size.toString(),
+          },
+          body: blob,
+        });
```
floryst, line 470: I'd like to handle loading, success, and failure states here. We can show a loading notification when uploading, then a success or failure depending on the state of the fetch.

src/components/App.vue
```diff
@@ -453,11 +460,49 @@ export default defineComponent({
...
+          });
+
+          if (saveResult.ok) messageStore.addSuccess('Save Successful');
+          else messageStore.addError('Save Failed', 'Network response not OK');
+        } catch (error) {
+          messageStore.addError(
+            'Save Failed with error',
+            `Failed from: ${error}`
+          );
+          console.error(error);
```
floryst, line 490: Don't need this console.error.

floryst:
LGTM!

PR #298, opened 2023-04-08
fix(io): avoid error when loading a zip in a zip

PaulHax:
With nested zip files, the original zip file would get mixed in with the extracted files, resulting in an Error: cannot find reader for .zip

Test file with nested zip.
[zip-in-zip.zip](https://github.com/Kitware/VolView/files/11182269/zip-in-zip.zip)

Also, remove unused and buggy makeZip()

### DIFF
--- a/src/io/io.ts
+++ b/src/io/io.ts
@@ -1,6 +1,7 @@
 import { extensionToImageIO } from 'itk-wasm';
 import { extractFilesFromZip } from './zip';
-import { DatasetFile } from '../store/datasets-files';
+import { DatasetFile, ZipDatasetFile } from '../store/datasets-files';
+import { partition } from '../utils';
 
 export const ARCHIVE_FILE_TYPES = new Set(['zip', 'application/zip']);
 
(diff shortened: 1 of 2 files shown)


floryst:
I'm seeing an error when loading the same session file twice. Repro steps:
1. Load the zip-in-zip.zip dataset
2. Paint in any view
3. Save the session
4. Without reloading, try to load the saved session file (I just did drag-n-drop)

image

PaulHax:
> I'm seeing an error when loading the same session file twice.

Hmm.  Seems loading a session.volview.zip containing a dataset that's already loaded has been a problem since at least 80caf04e654ebe4a3e921f962d6768398a6da6c7

floryst:
Looks like it's actually just any zip file, painting not needed. Repro'd on the stable branch.

1. Load a zip file
2. Save the session
3. Load the session without refreshing the page

floryst:
This PR looks good, so I'll open an issue for this other bug.

PR #300, opened 2023-04-10
fix(volume-coloring): opacity range is image range

floryst:
Do not use the opacity function's range. Instead, rescale it to the dataset range.


floryst:
Closing and re-creating to trigger a deploy preview.

PR #308, opened 2023-04-18
Rectangle Annotation Tool

PaulHax:
New annotation tool: Rectangle 

image

Adds `useAnnotationTool` for base functionality for ruler and rectangle tools.


floryst:
We definitely have logic we can make reusable generally for 2D annotations, e.g. `jumpTo`, getting the current frame of reference, context menu, etc. We can collaborate to make those changes; I want to separate them from the current ruler PR, since that one is getting a bit big.

PaulHax:
Yes, separate for sure.  I'm factoring out the `rulerStore` logic into a composable now.  Will definitly need your advice to get it right.

src/store/tools/useAnnotationTool.ts
```diff
@@ -0,0 +1,162 @@
...
+import { getLPSAxisFromDir } from '@/src/utils/lps';
+import { useViewConfigStore } from '@/src/store/view-configs';
+import { LPSAxisDir } from '@/src/types/lps';
+import { findImageID, getDataID } from '@/src/store/datasets';
+
+const useDependencyInjectionStore = defineStore('dependencyInjection', {});
+const useIDGenerator = () => {
+  const { $id } = useDependencyInjectionStore();
+  return $id;
+};
```
floryst, line 21: I think we should pass `this` to `useAnnotationTool`, which will provide us with the injected dependencies.

src/store/tools/useAnnotationTool.ts
```diff
@@ -0,0 +1,162 @@
...
+
+type AnnotationTool = {
+  id: unknown;
```
floryst, line 24: Is it possible to type this to `string`? `RectangleID extends string` should be true with that change, and same with `Rectangle extends AnnotationTool`. This means we can avoid the `ToolID extends Tool['id'] & string` below.

floryst:
Overall, looks good! I'm interested in also splitting out the `*Tool.vue` and `*Widget2D.vue` into composables as well. If you would rather tackle that at a later time, then that's fine as well.

I'm guessing the patch failures are due line number changes. Once I get the relevant vtk.js PRs in, we can drop the patch file altogether.

src/store/tools/useAnnotationTool.ts
PaulHax: How do I get a reference to `this` in the "setup" function of a store?
```
export const useRulerStore = defineStore('ruler', () => {
  this // nope
```

src/store/tools/useAnnotationTool.ts
PaulHax: Good idea.  Fixed.

PaulHax:
> Overall, looks good! I'm interested in also splitting out the *Tool.vue and *Widget2D.vue into composables as well. If you would rather tackle that at a later time, then that's fine as well. We don't have to do it all in this PR in the interest of time.

Sounds good.  Let me add Rectangles to the Annotations tab and this should be good to go and unblock that Vue 3 hotness.

src/store/tools/useAnnotationTool.ts
floryst: You can use `type _This = ReturnType<typeof useRulerStore>`, with `useAnnotationTool` using `Store` (named import from `pinia`) as the type.

PaulHax:
Retry with clear cache fixes Netlify build.
image

src/store/tools/useAnnotationTool.ts
PaulHax: I'm missing how to do this.  `useAnnotationTool` is in the top level "setup" arrow function scope where `this` is `undefined`.

```
export const useRectangleStore = defineStore('rectangles', () => {
  const toolStoreProps = useAnnotationTool({
    services: this, // this undefined in arrow function
    toolDefaults: rectangleDefaults,
  });
```

floryst: Oh, right! I forgot about the context. For function-based stores, `this` is only accessible on functions exposed on the store.

```
const useStore = defineStore('store', () => {
  type _This = Store // or type _This = ReturnType<typeof useStore>
  function generateID(this: _This) {
    return this.$id.nextID()
  }
  return {
    generateID,
  }
})
```

floryst: Let's just switch to an ID store. I forget why I opted to have ID generation in an injected dependency, but it probably doesn't apply anymore.

You won't need to pass in the ID generator to `useAnnotationTool`. You'd just call `useIdStore` in `useAnnotationTool`.

PaulHax: OK.  I added an IDStore and use it in composible.

src/store/tools/id.ts
```diff
@@ -0,0 +1,13 @@
+import { defineStore } from 'pinia';
+
+export const useIDStore = defineStore('IDStore', () => {
+  type _This = ReturnType<typeof useIDStore>;
+
+  function nextID(this: _This) {
+    return this.$id.nextID();
```
floryst, line 7: I was actually suggesting we get rid of the `this.$id` injection and just use an ID store, but I think that's out-of-scope. I can follow up with the changes necessary to accomplish this.

src/store/tools/useAnnotationTool.ts
```diff
@@ -54,10 +48,10 @@ export const useAnnotationTool = <Tool extends AnnotationTool>({
     return TOOL_COLORS[colorIndex];
   }
 
-  const idGenerator = useIDGenerator();
+  const idStore = useIDStore();
 
   const addTool = (tool: ToolPatch): ToolID => {
-    const id = idGenerator.nextID() as ToolID;
+    const id = idStore.nextID() as ToolID;
```
floryst, line 54: If you want to access `this.$id` in `addTool`, you'd need to change this to be the following. It works because `addTool` gets exposed as a store method, and pinia will go through and bind `this` to each exposed store method.

```
function addTool(this: _This, tool: ToolPatch): ToolID {
  const id = this.$id.nextID() as ToolID;
```

src/store/tools/rulers.ts
```diff
@@ -62,105 +49,22 @@ export const useRulerStore = defineStore('ruler', () => {
...
-      .forEach((ruler) => {
-        addRuler.call(this, ruler);
-      });
+  function deserialize(manifest: Manifest, dataIDMap: Record<string, string>) {
+    deserializeTools(manifest.tools.rulers, dataIDMap);
   }
 
   return {
-    rulers,
+    ...toolStoreProps,
```
floryst, line 63: Do we still need `toolStoreProps` spread here? Is there anything that is used?

floryst: Oh, I see `activateTool` and `deactivateTool` now. Let's just assign them, since we are polluting the exposed API here with stuff like `addTool`, which is aliased as `addRuler`.

floryst:
Overall, LGTM!

src/store/tools/rulers.ts
PaulHax: Another half baked thing there.  Would be nice to have a polymorphic toolStore shape no?  Makes Measurement.vue potentially nice dealing with single collection of tools (but the branded RectangleID type is sort of messing it up at the moment).  

So I will bite the bullet now and change all uses of `rulerStore.addRuler/updateRuler/etc` to `rulerStore.addTool/updateTool/etcTool`

floryst: I actually changed `MeasurementList` in the vue3 branch to be specific to the tool, since not all tools report the same info. So I have a `MeasurementRulerList` right now.

floryst: Let's just keep `addRuler/etc` for now. A general toolStore shape would be nice, but I want to be cautious in jumping into that first before we've had a chance to add more tools that aren't necessarily super similar to each other.

PaulHax: OK, I'll assign `activateTool` and `deactivateTool`, not spread the wrapped props and adjust `MeasurementList`

Issue #307, opened 2023-04-17
Loading sessions with zipped datasets fails


floryst:
Repro:

1. Load a zip file
2. Save the session
3. Load the session without refreshing the page

PR #314, opened 2023-04-27
Pipeline data loading

floryst:
- Data sources are imported via a flexible pipeline mechanism.
- Files are typed based on mime type rather than extension. Fixes a bug
where we do not handle the OS giving a mimetype in File.type.
- Magic handling and mime types are split out into their own files.

The bulk of the data source loading pipeline resides in `importDataSources.ts`.

### DIFF
--- a/src/core/importDataSources.ts
+++ b/src/core/importDataSources.ts
@@ -190,7 +190,7 @@ const extractArchive: ImportHandler = async (dataSource, { execute, done }) => {
           fileType: '',
         },
         archiveSrc: {
-          path: entry.archivePath,
+          path: `${entry.archivePath}/${entry.file.name}`,
         },
         parent: dataSource,
       });
--- a/src/core/importDataSources.ts
+++ b/src/core/importDataSources.ts
@@ -10,6 +10,7 @@ import { canFetchUrl, fetchFile } from '../utils/fetch';
 import { useImageStore } from '../store/datasets-images';
 import {
   DatasetFile,
+  isRemote,
   makeLocal,
   makeRemote,
   useFileStore,
@@ -107,6 +108,45 @@ export function convertDataSourceToDatasetFile(
 
   // local file case
   return makeLocal(fileDataSource.fileSrc!.file);
+}
+
+/**
+ * Converts a DatasetFile to a DataSource.
+ *
+ * This is an imperfect conversion, since DatasetFile contains less information
+ * (e.g. hierarchical provenance info).
+ * @param dataset
+ * @returns
+ */
+export function convertDatasetFileToDataSource(
+  dataset: DatasetFile,
+  options?: {
+    forceRemoteOnly: boolean;
+  }
+): DataSource {
+  const forceRemoteOnly = options?.forceRemoteOnly ?? false;
+  let parent: DataSource | undefined;
+
+  if (isRemote(dataset)) {
+    parent = {
+      uriSrc: {
+        uri: dataset.url,
+        name: dataset.remoteFilename,
+      },
+    };
+
+    if (forceRemoteOnly) {
+      return parent;
+    }
+  }
+
+  return {
+    fileSrc: {
+      file: dataset.file,
+      fileType: '',
+    },
+    parent,
+  };
 }
 
 function isArchive(r: DataSource): r is DataSource & { fileSrc: FileSource } {
--- a/src/core/importDataSources.ts
+++ b/src/core/importDataSources.ts
@@ -40,7 +40,7 @@ export interface FileSource {
 /**
  * If an archive source is specified, then it is assumed that the data source
  * has a FileSource (representing the file inside the archive), and a parent
- * data source that has a FileSource of the containing archive.
+ * data source that refers to the archive.
  */
 export interface ArchiveSource {
   // Full path + filename inside the archive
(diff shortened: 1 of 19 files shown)


src/core/importDataSources.ts
```diff
@@ -0,0 +1,190 @@
...
+  let src = dataSource;
+  const { fileSrc } = src;
+  if (fileSrc && fileSrc.fileType === '') {
+    const mime = await getFileMimeType(fileSrc.file);
+    if (mime) {
+      src = {
+        ...src,
+        fileSrc: {
+          ...fileSrc,
+          fileType: mime,
```
PaulHax, line 64: Do we want to assign the `File.type` with the found mime string and remove this property?

src/core/importDataSources.ts
```diff
@@ -0,0 +1,190 @@
...
+          fileType: '',
+        },
+        archiveSrc: {
+          path: entry.archivePath,
+        },
+        parent: dataSource,
+      });
+    });
+  } else {
+    next();
```
PaulHax, line 113: Should `next` just be the returned value?  Then normal functions can be middleware and encourage `return` for control flow rather than logic nesting.

Would that mean we need "sink" middleware to call a `endExecution()` func to avoid pushing undefined around?  And promise resolving in the middleware glue?

floryst: That should be doable. If a handler were to return rather than call next, then we'd need to detect undefined, which is fine. Nobody should be transforming data to undefined.
- Returning undefined means terminate the pipeline
- else, continue with that input

We would have to disable the eslint rule that requires consistent returns in all code paths, otherwise we'd have odd-looking `return undefined` everywhere.

```
const handler = (input, { execute }) => {
  if (isArchive(input)) {
    execute(extract(input))
    // can also disable eslint consistent returns
    return undefined
  }
  return input
}
```

Related, I am playing with having pipelines optionally return data as part of the pipeline result. This would mean that handlers must call `done(outputData)` if they want to terminate and produce data. Do you suppose it would make more sense for a non-undefined return value to designate the return value of a pipeline, or a transformed output of the handler?

src/core/importDataSources.ts
floryst: Probably, if we know we won't care to change `file.type` more than once. Assigning `fileType` is much easier than `file.type`, which is readonly and requires doing `file = new File([file], file.name, { type: newType })`. Having `fileType` just makes things easier, even if it's a redundant property slot.

src/core/importDataSources.ts
floryst: Here is a simpler approach. We will keep your idea of having handlers return the (optionally transformed, optionally wrapped in a Promise) input, and change the usage of `done()`. Calling `done()` signals to the pipeline that, when the handler returns, the pipeline should be terminated and the return value should be discarded.

```
const handler = (input, { execute, done }) => {
  if (isArchive(input)) {
    execute(await extract(input))
    // terminate processing of input
    done()
  }
  return input
}
```

PaulHax: Explicit `done()` makes sense.  

Just musing: another semantic option which makes sense for this particular unwrapping, recursive use case: `execute()` is always done with returned values.  But opting into recursion makes more sense in the general case probably.

floryst: I would rather have `execute()` run with explicit arguments, since it's more obvious. Middleware return values are strictly to pass along to the next pipeline handler. It's possible that a handler wants to execute a nested pipeline _and_ pass control along to the next handler.

src/core/importDataSources.ts
floryst: I actually have a a possible reason to not re-assign the `file` property: it might be the case that we care about matching pipeline return values with File inputs by comparing the File instance. I'm opting to keep the `fileType` property for now.

floryst:
The latest commit simplifies the handler API and supports pipeline return values via `done(returnValue)`.

floryst:
The pipeline importer is now fully integrated with the most recent changes!

`dataset.loadFiles` now handles translating the inputs and outputs to `importDataSources`, leaving `importDataSources` to do the bulk of the work loading datasets.

floryst:
I've changed this to target `main` rather than `vue3`, since it rebases cleanly.

PaulHax:
Error on loading a session.volview.zip because the `extractFilesFromZip` returns filename in path now and this check is coming up empty:
https://github.com/Kitware/VolView/blob/44437af3dd7bcf4e8f94fbab22c6f2c56f40d8d9/src/io/state-file/utils.ts#L107-L110

floryst:
Fixed the session loading bug by reverting my changes. Also added back the `Maybe` type.

floryst:
Fixed more session-related bugs.

PaulHax:
Got some errors with session loading of remote dataset.  Tried saving/loading prostate.  Here are some hacks pointing to issues:
https://github.com/Kitware/VolView/pull/317

floryst:
I pushed a new commit that allow for the prostate example to be saved and loaded back.

PaulHax:
Squashed it.  🍨

But another!  State file loaded from URL parameter gets treated as a normal zip.  It's not downloaded and resolved as a zip yet before this check:
```
    // For now only support restoring from a single state files.
    stateFile = files.length === 1 && (await isStateFile(files[0].file));
    if (stateFile) {
      statuses = await loadState(files[0].file);
    } else {
      statuses = await dataStore.loadFiles(files);
    }
```

PaulHax (approved):
Works good.  Code, much better!

Closes #307 ?

PR #315, opened 2023-04-28
DICOMWeb fixes for Google hosted service

PaulHax:
- Remove trailing slash if on host URL
- Support minimal metadata responses
- Add a `dicomweb=http://localhost:5173/dicom-web` URL parameter to configure default host

### DIFF
--- a/src/store/dicom-web/dicom-web-store.ts
+++ b/src/store/dicom-web/dicom-web-store.ts
@@ -102,7 +102,7 @@ export const useDicomWebStore = defineStore('dicom-web', () => {
       seriesInstanceUID: volumeInfo.SeriesInstanceUID,
       sopInstanceUID: middleInstance.SopInstanceUID,
     };
-    return fetchInstanceThumbnail(cleanHost.value!, instance);
+    return fetchInstanceThumbnail(cleanHost.value, instance);
   };
 
   const fetchPatientMeta = async (patientKey: string) => {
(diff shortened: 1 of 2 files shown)


src/store/dicom-web/dicom-web-store.ts
```diff
@@ -43,10 +45,17 @@ async function getAllPatients(host: string): Promise<PatientInfo[]> {
...
-    ? ref(process.env.VUE_APP_DICOM_WEB_URL)
-    : useLocalStorage<string>('dicomWebHost', '');
-  const isConfigured = computed(() => !!host.value);
+  const urlParams = vtkURLExtract.extractURLParameters() as UrlParams;
+  const dicomWebFromURLParam = urlParams.dicomweb as string | undefined;
+  const initialHost = dicomWebFromURLParam ?? process.env.VUE_APP_DICOM_WEB_URL;
+
+  const host = initialHost
+    ? ref(initialHost)
+    : useLocalStorage<string | null>('dicomWebHost', ''); // null if cleared by vuetify text input
```
floryst, line 54: Should `initialHost` also be saved to local storage, or do you think the `dicomweb` parameter should be an overriding exception?

src/store/dicom-web/dicom-web-store.ts
```diff
@@ -43,10 +45,17 @@ async function getAllPatients(host: string): Promise<PatientInfo[]> {
  * Collect DICOM data from DICOMWeb
  */
 export const useDicomWebStore = defineStore('dicom-web', () => {
-  const host = process.env.VUE_APP_DICOM_WEB_URL
-    ? ref(process.env.VUE_APP_DICOM_WEB_URL)
-    : useLocalStorage<string>('dicomWebHost', '');
-  const isConfigured = computed(() => !!host.value);
+  const urlParams = vtkURLExtract.extractURLParameters() as UrlParams;
+  const dicomWebFromURLParam = urlParams.dicomweb as string | undefined;
```
floryst, line 49: I'd recommend moving `dicomweb` into a constant. It makes it more obvious if we need to ever configure it down the road.

PaulHax: Done.

floryst:
LGTM!

PR #320, opened 2023-05-04
fix(io): Remove private qualifier initialize method in DICOMIO Class

YeonHoJeong:
The 'initialize()' method, which was declared with a 'private' qualifier inside the DICOMIO class, is being called from outside and thus the 'private' qualifier is removed.


floryst:
Thanks for your contribution!

PR #323, opened 2023-05-09
Labels for annotation tools 

PaulHax:
Adds a config JSON file processor that sets the available labels for ruler and rectangle tools.

Adds MenuToolButton component for the ToolStrip.  Paint color and label selection menus pop left and stay open until tool button is clicked again. 

Adds `q` and `w` keyboard shortcuts to cycle through active label selection when label menu is open.


floryst:
Here are some of my thoughts so far on labels. Having a single label store will likely result in excess properties down the road. For instance, we may add a "lineStrokeStyle" property for rulers and rectangles. However, that would not make any sense for labels used for painting. In general, there will be properties that apply to one tool but not necessarily the others.

It may make more sense for each tool store to have their own labels + label properties. I'm thinking along the lines of a label store composable, similar to what you did with useAnnotationTools. The labels API is straightforward, and all the composable needs to do is create/get/update/delete labels and their properties. Tools can then have their own labels with their own specific properties.

floryst:
Just checking in on this. Is this still a draft?

PaulHax:
Still needs work.  Almost re-based on vue3 goodness.  

Then could be good time to use ideas from your `vue3-improve-toolbutton` branch.  

- [ ] Refactor this branches' LabelToolButton to a MenuToolButton that takes a slot with menu contents.
- [ ] Pull out label selection radio group from LabelToolButton and slot in for Ruler and Rectangle MenuToolButtons

Then get closer to your label doc vision.  Maybe punt on UI level create/delete labels for now but need to:

- [ ] Ensure that changed color/props of configured labels updates color/props of existing tools on deserialization

At the moment, each tool store has there own set of labels in its useLabels composable.  But can only configure labels globally with a single `labels={name: color, nameB: anotherColor}` URL parameter.    No `rulerLabels` and `rectangleLabels` URL parameters or somesuch.  Should we aim for that in this PR or save for latter?

Other things?

documentation/content/doc/toolbar.md
```diff
@@ -22,12 +22,22 @@ When the paint tool is selected, you can paint in any 2D window. Click on the pa
 
 When the ruler tool selected, the left mouse button is used to place and adjust ruler end-markers. Right clicking on a end-marker displays a pop-up menu for deleting that ruler. Switch to the "Annotations" tab to see a list of annotations made to currently loaded data. Select the location icon next to a listed ruler to jump to its slice. Select the trashcan to delete that ruler.
 
+Ruler annotations can be tagged with a label. Configure labels with
+the `labels` URL parameter. `labels` is a list with `name, color` tuples.
+
+https://volview.netlify.app/?labels=[artifact,888888,innocuous,00ff00]
```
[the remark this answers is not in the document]
PaulHax: I like that idea.  Perhaps this json should just be the session manifest.json?  That way we just maintain 2 "config" interfaces, URL params and session files.

src/components/ToolStrip.vue
```diff
@@ -53,83 +49,61 @@
     </groupable-item>
     <div class="my-1 tool-separator" />
     <groupable-item v-slot:default="{ active, toggle }" :value="Tools.Paint">
-      <v-menu
-        v-model="paintMenu"
-        location="end"
-        :close-on-content-click="false"
-        :disabled="!active"
+      <MenuToolButton
```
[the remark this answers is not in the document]
PaulHax: Yes!

src/components/MenuToolButton.vue
```diff
@@ -0,0 +1,93 @@
+<script lang="ts">
+import { watch, ref, computed, defineComponent } from 'vue';
+import { useDisplay } from 'vuetify/lib/framework.mjs';
+import ToolButton from './ToolButton.vue';
+
+export default defineComponent({
+  name: 'LabelToolButton',
```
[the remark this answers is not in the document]
PaulHax: Derp

documentation/content/doc/toolbar.md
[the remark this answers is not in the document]
floryst: I think just the session manifest.json is a good starting point for URL-based configuration.

I don't understand what you mean by abstract labels. We should specify labels per tool for now, unless you had a different idea.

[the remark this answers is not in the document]
PaulHax: Added a file processor that takes a JSON to configure labels.

(The pipeline stuff is pretty slick.)

src/io/import/importDataSources.ts
```diff
@@ -60,6 +61,8 @@ export async function importDataSources(dataSources: DataSource[]) {
   const middleware: Array<ImportHandler> = [
     // updating the file type should be first in the pipeline
     updateFileMimeType,
+    // handleConfig must be for restoreStateFile for label props to be applied to deserialized tools
```
floryst, line 64: I assume you meant to say "must be before restoreStateFile".

src/io/import/processors/handleConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+const readConfigFile = async (configFile: File) => {
+  const decoder = new TextDecoder();
+  const ab = await configFile.arrayBuffer();
+  const text = decoder.decode(new Uint8Array(ab));
+  return Config.parse(JSON.parse(text));
+};
+
+const applyConfig = (manifest: z.infer<typeof Config>) => {
+  useRectangleStore().setLabels(manifest.rectangleLabels);
+  useRulerStore().setLabels(manifest.labels);
```
floryst, line 34: Suggestion: `labels` should be `rulerLabels`. Unless this is meant to be a generic/abstract "labels" definition?

src/store/tools/useAnnotationTool.ts
```diff
@@ -1,34 +1,44 @@
...
+const annotationToolDefaults = {
+  frameOfReference: {
+    planeOrigin: [0, 0, 0],
+    planeNormal: [1, 0, 0],
+  },
+  slice: -1,
+  imageID: '',
+  placing: false,
+  color: TOOL_COLORS[0],
+  name: 'baseAnnotationTool',
```
floryst, line 27: Should we have a default name? Or just leave it empty?

src/store/tools/useLabels.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useLabels = <Tool>(initialLabels: Labels<Tool>) => {
+  const labels = ref(initialLabels);
+
+  const activeLabel = ref<string | undefined>();
+  const setActiveLabel = (name: string) => {
+    activeLabel.value = name;
+  };
+
+  const setLabels = (newLabels: Labels<Tool> | null | undefined) => {
```
floryst, line 14: FYI don't forget `Maybe<Labels<Tool>>` will simplify `| undefined` and `| null | undefined` for cases where we don't care about null/undefined distinction.

documentation/content/doc/toolbar.md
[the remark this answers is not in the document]
floryst: We'll need to update the docs once we hammer out the config details.

src/io/import/processors/handleConfig.ts
PaulHax: I was thinking it could be the generic fallback "labels" but I did not implement. Now I did like this:
```
const applyConfig = (manifest: z.infer<typeof Config>) => {
  useRulerStore().setLabels(manifest.labels);
  useRectangleStore().setLabels(manifest.rectangleLabels ?? manifest.labels);
};
```

src/store/tools/useAnnotationTool.ts
PaulHax: no opinion.  Was just following ruler which has 'Ruler" as default name.

src/store/tools/useLabels.ts
PaulHax: Oh, yeah!

documentation/content/doc/toolbar.md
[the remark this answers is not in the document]
PaulHax: Docs are updated with a config.json example.

src/config.ts
```diff
@@ -167,14 +167,29 @@ export const SAMPLE_DATA: SampleDataset[] = [
...
+  red: { color: 'red' },
+  green: { color: '#00ff00' },
+  gray: { color: '#888888' },
+};
+
+export const RECTANGLE_LABEL_DEFAULTS = {
+  lesion: { color: 'red', fillColor: 'transparent' },
+  innocuous: { color: '#00ff00', fillColor: '#00ff0020' },
+  artifact: { color: '#888888', fillColor: '#ffbf0040' },
+};
```
floryst, line 188: @aylward any suggestions for defaults?

src/io/import/processors/handleConfig.ts
floryst: In that case, I would also suggest allowing for a `rulerLabels` as well, since `labels` is the default fallback and each other tool can have their own labels.

floryst (commented):
Looking good overall!

src/components/MenuToolButton.vue
```diff
@@ -0,0 +1,92 @@
...
+});
+</script>
+
+<template>
+  <v-menu
+    persistent
+    no-click-animation
+    :close-on-content-click="false"
+    v-model="menuOn"
+    :location="showLeft ? 'left' : 'right'"
```
floryst, line 49: @aylward what's your opinion on these menus (e.g. paint menu) showing up on the left vs the right?

@PaulHax if we opt for left-opening menus, we need to make sure the menu content has enough [elevation](https://vuetifyjs.com/en/styles/elevation/). Right now, left menus blend in with the sidebar.

src/components/MenuToolButton.vue
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    return { showLeft, menuOn, handleActivatorClicked };
+  },
+});
+</script>
+
+<template>
+  <v-menu
+    persistent
```
floryst, line 45: Any reason to make this persistent? I think tool menus don't have a reason to be persistent, especially since we disable close-on-content-click.

PaulHax: Persistent keeps the menu open after making an annotation.  Saves a click if you want to make a new annotation with a new label next.

floryst: While I agree that is useful, I think it goes against users' expectations for how menus operate, and it's not immediately obvious that clicking on the tool again will toggle the menu. The tool menu operates more like a context menu, rather than a persistent menu. Do we have user feedback on having this menu be persistent?

My suggestion would be to add support for controlling tools from the Annotations sidebar, which we can do in a later PR. That would allow users to not have to add an additional click to switch between labels.

src/components/MenuToolButton.vue
PaulHax: Added `elevation-24` on the menu popup.

src/components/MenuToolButton.vue
PaulHax: Makes total sense.  I can work on that next week.

src/io/import/processors/handleConfig.ts
PaulHax: Added the `rulerLabels` property for the config.

src/components/MenuToolButton.vue
PaulHax: I removed the persistent attribute.

PR #332, opened 2023-05-16
refactor(ToolButton): improve API and support menus

floryst:
Adds an easier-to-use API and menu support to cut down on duplicate markup. Additonally makes menu more discoverable by popping up on first click..

FYI @PaulHax this is going into the vue3 branch. I want to keep this separate since it's an added internal feature.

### DIFF
--- a/src/components/PaintControls.vue
+++ b/src/components/PaintControls.vue
@@ -77,6 +77,7 @@
             hide-mode-switch
             hide-sliders
             show-swatches
+            elevation="0"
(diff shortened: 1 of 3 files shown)


PaulHax:
I did the same on label branch; first thing one wants to do after switching tools, pick a color/abel.

floryst:
Nice! We can work together to merge the two as necessary. I haven't thoroughly looked through your changes, other than seeing that you made a LabelToolButton component.

PR #334, opened 2023-05-24
fix(WindowLevelTool): swap W/L controls

floryst:
Swaps Window and Level controls.

### DIFF
--- a/src/components/tools/WindowLevelTool.vue
+++ b/src/components/tools/WindowLevelTool.vue
@@ -82,17 +82,17 @@ const WindowLevelToolComponent = defineComponent({
     const vertVal = ref(0);
     const horizVal = ref(0);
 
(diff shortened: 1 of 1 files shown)


PaulHax (approved):
I always wondered what was the convention.

When I drag the mouse down, the image gets darker.  Should the level direction be inverted?

floryst:
I'm not 100% sure. I need to check to see if it's consistent across other viewers.

PaulHax:
Level decreases with mouse moving down, matching current behavior:
* [NilRead ](https://www.hyland.com/en/products/medical-imaging/nilread)
* 3D Slicer
 
Increases the level with the mouse moving down.  Opposite of current behavior.
* OHIF

floryst:
Not planned at the moment.

PR #335, opened 2023-05-24
fix(SaveSession): loading state and no refresh

floryst:
- Saving now shows a loading indicator.
- Prevent refresh when using Enter to save the session.


PaulHax (approved):
Verks good!

PR #340, opened 2023-06-07
Label Controls on Annotation tab

### LINKED ISSUE #339: Tool label selection on Annotations tab
Annotators need a quick way to switch between active tool labels.  

Add a section to the Annotations tab that shows the labels for the active tool.  Something like this and using [ItemGroup](https://vuetifyjs.com/en/components/item-groups/)

labels

When the user clicks the Ruler or Rectangle Tool, automatically switch to the Annotations tab.

PaulHax:
Adds Tool Controls to Annotation Tab for selected tool (just ruler and rectangle).  

Activating ruler or rectangle tools automatically switches to Annotations tab.

Tool Controls show a label selection UI.  

Layout of measurement lists got tweaked.  Added label for each  measurement item as requested.

image

closes #339


PaulHax:
Auto switching to the Annotations tab upon tool activation is... something.  Kind of make the popup menu next to the tool button redundant.

floryst:
> Auto switching to the Annotations tab upon tool activation is... something. Kind of makes the popup menu next to the tool button redundant (at least in non-mobile layout mode.)

If we think the annotations panel is a better workflow for users, I would vote for removing the popup menu entirely.

src/components/AnnotationsModule.vue
```diff
@@ -52,3 +46,15 @@ export default defineComponent({
   display: block;
 }
 </style>
+
+<style>
+.annotation-header {
```
floryst, line 51: I would recommend moving this out into its own css stylesheet in `components/styles` and doing a `<style scoped src="..." />` to keep names scoped.

src/composables/useKeyboardShortcuts.ts
```diff
@@ -0,0 +1,48 @@
...
+  const handleKeyDown = (event: KeyboardEvent) => {
+    handleLabelSection(event);
+  };
+
+  onMounted(() => {
+    window.addEventListener('keydown', handleKeyDown);
+  });
+  onUnmounted(() => {
+    window.removeEventListener('keydown', handleKeyDown);
+  });
```
floryst, line 47: Just a note: you can use [`onKeyStroke`](https://vueuse.org/core/onKeyStroke/) if it simplifies things for you. Otherwise we can leave this as-is.

Just some sketching below.

```
onKeyStroke(DECREMENT_LABEL_KEY, (e) => applyLabelOffset(-1))
```

src/components/ModulePanel.vue
```diff
@@ -68,6 +70,15 @@ export default defineComponent({
   setup() {
     const selectedModuleIndex = ref(0);
 
+    const toolStore = useToolStore();
+    watch(
+      () => toolStore.currentTool,
+      (newTool) => {
+        if ([Tools.Rectangle, Tools.Ruler].includes(newTool))
```
floryst, line 77: I would recommend making this array a module-level constant for visibility purposes.

src/components/ToolControls.vue
```diff
@@ -0,0 +1,25 @@
...
+  [Tools.Rectangle, { component: RectangleControls, label: 'Rectangle' }],
+  [Tools.Ruler, { component: RulerControls, label: 'Ruler' }],
+]);
+
+const tool = computed(() => tools.get(toolStore.currentTool));
+</script>
+
+<template>
+  <div v-if="tool">
+    <div class="annotation-header">{{ tool.label }}</div>
```
floryst, line 20: Use `<style scoped src="..." />` to keep `annotation-header` scoped. (See comment above in AnnotationsModule.)

floryst:
Some usage comments:
- I think we should disable the color picker control. Right now, we can click on the colored circle next to a measurement and change its color, which I don't think is desirable if we have labels. We can keep the colored circle, but just disable the color picker popup.
- I'm not sure about prefixing the measurement name with "Label: ". Do you think it makes more sense to explicitly prefix with "Label:" or just have the label name be the measurement name?
- Right now we don't have a good way of associating a measurement entry with the measurement in the scene. We should brainstorm how to do this for a future PR.
- I'm liking the label controls! Some styles to play around with:
  - circles instead of squares (since we are using rounded items)
  - denser item buttons. It seems too spaced out right now.

floryst:
That sounds like a good idea!

floryst:
LGTM!

PR #346, opened 2023-06-12
Sentry

PaulHax:


.env.example
```diff
@@ -2,4 +2,8 @@
 # If no variable found, DICOMWeb panel in data tab is hidden unless set in settings.
 VUE_APP_DICOM_WEB_URL=http://localhost:5173/dicom-web
 # Display name of DICOMWeb host
-VUE_APP_DICOM_WEB_NAME=ACME Hospital
\ No newline at end of file
+VUE_APP_DICOM_WEB_NAME=ACME Hospital
+
+VUE_APP_SENTRY_DSN=xhttps://330d4ca81b9c48cfac837d395966e6ee@o4505347877765120.ingest.sentry.io/4505347880779776
+
+VUE_APP_ENABLE_REMOTE_SAVE=true
```
floryst, line 9: I like this idea generally. There are two thoughts here:
- builds must explicitly set this to true
- the save dialog puts up a warning if there is a saveUrl

By the way, I don't see this env var used anywhere in this PR.

floryst: Got it. I must have missed it before.

Let's also add a few comments describing what each env var does.

floryst:
I like the general idea of putting it on the landing page. The button feels a bit big. Maybe a dense button?

PaulHax:
Button is smaller now:
image

aylward:
When I run the version 
   https://deploy-preview-346--volview-dev.netlify.app/
the button doesn't appear on the landing page, and I don't have an option to enable (opt-in) in the settings.

Once I can test, I'd like to provide more comments.

aylward:
Should (can) we hard-code the sentry info, instead of storing it in a config file?   Seems like it isn't something that should change unless the main app is being changed in some way, or am I missing a common use case?

.env.example
```diff
@@ -1,5 +1,13 @@
...
 # If no variable found, DICOMWeb panel in data tab is hidden unless set in settings.
 VUE_APP_DICOM_WEB_URL=http://localhost:5173/dicom-web
+
 # Display name of DICOMWeb host
-VUE_APP_DICOM_WEB_NAME=ACME Hospital
\ No newline at end of file
+VUE_APP_DICOM_WEB_NAME=ACME Hospital
+
+# Error reporting URL for sentry.io.  Each sentry project has its own DSN.
+VUE_APP_SENTRY_DSN=xhttps://330d4ca81b9c48cfac837d395966e6ee@o4505347877765120.ingest.sentry.io/4505347880779776
```
floryst, line 9: I think we should leave this empty for the example file.

floryst:
Only our public builds will include our sentry logger. We don't want to enable sentry for other projects that build VolView for their own needs.

aylward:
Ah.  Makes sense.

floryst:
Last thing: can we hide the "opt out of error reporting" when error reporting is already off?

PR #348, opened 2023-06-15
Migrate to vite

floryst:
- Migrates to use vite tooling
- Adds a bundle size analyzer
- Reduces bundle size
- ~add unit test runner~


floryst:
This actually needs a bit more work to get unit tests back.

floryst:
I've removed unit testing for now. I'll follow up with a new PR for testing to avoid mixing things up, since I want to enable running unit tests in CI.

floryst:
This also changes the default colormap to CT-AAA @aylward

PR #350, opened 2023-06-16
Add vitest

floryst:
- creates a `useIdStore()` instead of `this.$id`, since there's a type conflict with pinia's `this.$id`.
- adds vitest
- adjusts tests to pass
- adds testing to CI


PaulHax (approved):
Cool!

PR #352, opened 2023-06-21
fix : modify slot name in ViewOverlayGrid

YeonHoJeong:
the name of the slot looks awkward in ViewOverlayGrid


floryst:
Looks good to me!

PR #356, opened 2023-06-24
Vite migration tweeks

PaulHax:
* Fix layers by passing runtime pipeline paths to itk-wasm
* Move index.html to root to avoid CLI warning.


floryst:
Good catch. LGTM!

PR #358, opened 2023-06-27
Tool Labels: Add, Edit, Delete

### LINKED ISSUE #347: Annotation Tool Labels: Create, Edit, Delete
## UI

Create an "Add Label" chip button at end of LabelControls.  Clicking "Add Label" pops "Edit Label" modal.

Edit Label modal form:
* Label name
* Properties (color, fill color, etc)
* Delete button

Add "Edit Label" icon button to each LabelControl chip which pops modal.

After label is edited, existing annotations with label have their properties updated.  Annotations with a deleted label get their label set to empty string. 

Change label for existing annotation is for another issue.

## Implementation

Serialization: Each annotation tool saves all their labels + properties, including labels from the config.json.  On deserialization, session manifest.json labels are merged with config.json labels, with manifest.json labels taking precedence.

PaulHax:
Adds edit label and add label buttons to label list card:
image

Clicking the pencil button pops a modal:
image

closes #347

### DIFF
--- a/src/io/state-file/schema.ts
+++ b/src/io/state-file/schema.ts
@@ -12,6 +12,7 @@ import { Rectangle } from '@/src/types/rectangle';
 import { LPSCroppingPlanes } from '@/src/types/crop';
 import { FrameOfReference } from '@/src/utils/frameOfReference';
 import { Optional } from '@/src/types';
+import { AnnotationTool } from '@/src/types/annotationTool';
 
 import {
   CameraConfig,
@@ -252,9 +253,7 @@ const FrameOfReference: z.ZodType<FrameOfReference> = z.object({
   planeNormal: Vector3,
 });
 
-const Ruler: z.ZodType<Ruler> = z.object({
-  firstPoint: Vector3,
-  secondPoint: Vector3,
+const AnnotationTool = z.object({
   imageID: z.string(),
   frameOfReference: FrameOfReference,
   slice: z.number(),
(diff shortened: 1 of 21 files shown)


src/io/state-file/schema.ts
```diff
@@ -252,30 +253,33 @@ const FrameOfReference: z.ZodType<FrameOfReference> = z.object({
...
+  z.object({ tools: z.array(tool), labels: z.record(tool.partial()) });
 
-const Rectangle: z.ZodType<Optional<Rectangle, 'fillColor'>> = z.object({
+const Ruler = AnnotationTool.extend({
   firstPoint: Vector3,
   secondPoint: Vector3,
-  imageID: z.string(),
-  frameOfReference: FrameOfReference,
-  slice: z.number(),
+}) satisfies z.ZodType<Ruler>;
```
PaulHax, line 273: `satisfies` seems to get the same type safety as the variable type declaration, while allowing me to `ZodObject.extend`.  Should I convert the rest of schema to use this pattern?

src/io/state-file/schema.ts
```diff
@@ -301,8 +305,8 @@ const LPSCroppingPlanes: z.ZodType<LPSCroppingPlanes> = z.object({
 const Cropping = z.record(LPSCroppingPlanes);
 
 const Tools = z.object({
-  rulers: Ruler.array(),
-  rectangles: Rectangle.array().optional(),
+  rulers: Rulers.optional(),
+  rectangles: Rectangles.optional(),
```
PaulHax, line 309: Lesson here is to always make big concept entries Objects, so can extend them with new props (like labels) without breaking backwards compatibility.

src/io/state-file/schema.ts
floryst: Yep! This pattern looks good to me.

src/utils/index.ts
```diff
@@ -254,3 +254,12 @@ export function cleanUndefined(obj: Object) {
     {}
   );
 }
+
+// converts named colors (red, antiquewhite, etc) to hex
+export function standardizeColor(color: string | undefined) {
```
floryst, line 259: Can use `Maybe<string>` here, since we don't differentiate between null and undefined.

src/store/tools/useAnnotationTool.ts
```diff
@@ -127,26 +142,45 @@ export const useAnnotationTool = <
...
+  };
+
+  type Serialized = {
+    tools: PartialWithRequired<Tool, 'imageID'>[];
+    labels: Labels<Tool>;
+  };
   function deserialize(
     this: Store,
-    serialized: PartialWithRequired<Tool, 'imageID'>[],
+    serialized: Serialized | undefined,
```
floryst, line 167: can use `Maybe<Serialized>` here, since we don't differentiate between null and undefined.

floryst:
Some UI suggestions:
- I think the edit button background doesn't look great given that it intersects with the label border. Lets try either removing the button background altogether or introduce a small gap between the button and the label border.
- I think we should have a "save/done" button in the label editor, even though all changes persist immediately. Not all users will understand that changes are auto-saved. Side-note: we may need to add a "cancel" button if users would rather changes not persist immediately, but I don't think we should handle that case right now.
- I think we may re-examine the layout of the label editor down the road. I like how it is currently because it allows for more fields in the future, but something still feels off. Maybe it's the lack of margins around the color picker and the text editor.

PaulHax:
> Some UI suggestions:

Good tips.  I removed the outline on the edit label button, added a "close" button, and introduced some white space to edit label module.

floryst (approved):
LGTM!

PR #359, opened 2023-06-29
New Feature: Multiplanar Oblique Views

jadh4v:
### Sub-tasks:
- [x] Add ResliceCursorWidget to control the plane normals
- [x] Merge related vtk-js PR
 https://github.com/Kitware/vtk-js/pull/2850
- [x] Add a global reset views button
- [x] Add 3D view to see all the cutting planes
- [x] Upgrade vtk-js after merging tsdefs:
       https://github.com/Kitware/vtk-js/pull/2906

### [Additional Fixes](#additional-fixes) (9/28/2023):
 - [x] Make image outline semi-transparent
 - [x] Change outline colors to less saturated / less bright
 - [x] Sync colors with orientation box
 - [x] 3D view - parallel projection
 - [x] Fix initial reset-camera by using a composite representation
 - [x] remove custom-events / replace with createEventHook
 - [x] gray out tool bar buttons that don't work with oblique views

### A complete list of all vtk-js PRs related to this Volview feature:
https://github.com/Kitware/vtk-js/pull/2908
https://github.com/Kitware/vtk-js/pull/2906
https://github.com/Kitware/vtk-js/pull/2896
https://github.com/Kitware/vtk-js/pull/2895
https://github.com/Kitware/vtk-js/pull/2888
https://github.com/Kitware/vtk-js/pull/2883
https://github.com/Kitware/vtk-js/pull/2850
https://github.com/Kitware/vtk-js/pull/2847

https://github.com/Kitware/VolView/assets/22624785/86b887db-e7ea-49a6-a8e1-5670fff86924


jadh4v:
@floryst 
Still some pending stuff, not yet ready for full review, but feel free to throw you comments on the current state of the code.

package.json
```diff
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.321.1",
-    "@kitware/vtk.js": "^27.4.6",
+    "@kitware/vtk.js": "../vtk-js/dist/esm",
```
jadh4v, line 26: Need to update this once the vtk-js branch is merged.

jadh4v:
@floryst @finetjul Please review

src/components/tools/ResliceCursorTool.vue
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    const state = resliceCursor.getWidgetState() as ResliceCursorWidgetState;
+    state.getStatesWithLabel('sphere').forEach((handle) => (handle as ResliceCursorWidgetState).setScale1(20));
+
+    const widget = ref<vtkResliceCursorViewWidget>();
+
+    const VTKViewType = computed(() => {
+      const viewAxis = getLPSAxisFromDir(viewDirection.value as LPSAxisDir);
+      switch(viewAxis) {
```
luciemac, line 64: It would be great to have this _switch case_ as a utility function !

src/components/LayoutGrid.vue
```diff
@@ -66,6 +74,52 @@ export default defineComponent({
...
+    resliceCursor.value = vtkResliceCursorWidget.newInstance();
+    // Orient the planes of the vtkResliceCursorWidget to the orientation
+    // of the currently set image.
+    const resliceCursorState = resliceCursor.value.getWidgetState() as ResliceCursorWidgetState;
+    provide(VTKResliceCursor, resliceCursor);
+
+    watch(currentImageData, (image) => {
+    // Reset to default plane values before transforming based on current image-data.
+    resliceCursorState.setPlanes({
+      4: {
```
luciemac, line 89: I think it is clearer to use `[ViewTypes.XY_Plane]`,  `[ViewTypes.YZ_Plane]`, `[ViewTypes.XZ_Plane]` syntax instead of 4, 5, 6.

src/components/VtkObliqueView.vue
```diff
@@ -0,0 +1,681 @@
...
+        <slice-scroll-tool :view-id="viewID" />
+        <window-level-tool :view-id="viewID" />
+      </div>
+      <view-overlay-grid class="overlay-no-events view-annotations">
+        <template v-slot:bottom-left>
+          <div class="annotation-cell">
+            <div>Slice: {{ currentSlice }}/{{ sliceMax }}</div>
+            <div
+              v-if="
+                typeof windowWidth === 'number' &&
```
luciemac, line 48: `windowWidth` seems to be defined as a `ComputedRef<Nullable<number>>`. Why do you need to check its type here? Can't you check it is not null?

floryst:
I seem to be getting some initial flickering behavior. Repro: load the protate example, then switch to the oblique only view. Moving my mouse over the three views (intersecting with the reslice cursor widget) causes the rendering to change.

jadh4v:
> I seem to be getting some initial flickering behavior. Repro: load the protate example, then switch to the oblique only view. Moving my mouse over the three views (intersecting with the reslice cursor widget) causes the rendering to change.

Updated the branch to fix this issue, and some other interaction issues regarding panning operation. Please test again.

aylward:
This is outstanding!  Very responsive on my slow/old laptop.   No memory issues or crashing!

Very Nice!

I think we need to tweak the look-and-feel a bit:

1. The control/indicator lines are too thick.  They obscure too much of the data.
2. The circles for controlling rotation also cover too much data - probably better to move them towards the ends of the lines instead of near the middle where interesting things happen in the data.   
3. I do not think we should use a change in the cursor to indicate function - the item in the scene should clearly indicate function, without having to cursor over it.   So, the circle handles to indicate "grab these to rotate" should instead look like rotation widgets that are intuitive.
4. The rotation cursor is subtle - still primarily an arrow with a small rotation indicator added.  The entire cursor should change to a rotation indicator.
5. It requires mental effort to map which window changes when I move - that is, which frame does the blue line correspond to.  The windows need boundaries indicating which line they belong to.
6. We need a 3D view!   I got completely lost in the data when I started changing more than one axis, and I cannot figure out how to get it back to the original view.  The 3D view should show the 3 slices and the bounding box of the data.   It should also use color or some indicator to make it very intuitive/fast/low-effort map each 3D slice to the corresponding 2D frame.  I think we should use color borders like 3D Slicer in the oblique view - I cannot think of anything more intuitive.
7. We need a reset button - to go back to the original view.   I tried, but cannot figure out why, in the image below, the data continues to be tilted.   
image
Here is how the data looks (nicely square) when I first load it
image

Again - very nice!  Thank you for the contribution!!!

s

src/components/VtkObliqueView.vue
```diff
@@ -316,11 +316,32 @@ export default defineComponent({
...
+          break;
+        case 'Superior':
+        case 'Inferior':
+          obliqueRep.setOutlineColor([0, 0, 1]);
+          break;
+        default:
+          obliqueRep.setOutlineColor([0, 0, 0]);
+          break;
+      };
+    }
```
floryst, line 343: I recommend putting this inside a `watch(baseImageRep, (obliqueRep) => ...), { immediate: true })`.

src/components/VtkObliqueView.vue
```diff
@@ -0,0 +1,671 @@
...
+        resliceCursorRef.value.updateCameraPoints(viewProxy.value.getRenderer(), VTKViewType.value, true, false, true);
+      }
+
+      const onPlanesUpdated = useVTKCallback(
+        resliceCursorRef.value.getWidgetState().onModified
+      );
+
+      onPlanesUpdated(() => {
+        updateViewFromResliceCursor();
+      });
```
floryst, line 381: This should be moved outside of onMounted, so that the lifecycle methods in useVTKCallback work.

```
      const onPlanesUpdated = useVTKCallback(
        computed(() => resliceCursorRef.value?.getWidgetState().onModified)
      );
```

src/shims-vtk.d.ts
```diff
@@ -295,3 +246,49 @@ declare module '@kitware/vtk.js/Widgets/Widgets3D/ImageCroppingWidget' {
...
+declare module '@kitware/vtk.js/Proxy/Representations/GeometryRepresentationProxy' {
+  // import vtkGeometryRepresentationProxy from '@kitware/vtk.js/Proxy/Representations/GeometryRepresentationProxy';
+  import vtkAbstractRepresentationProxy from '@kitware/vtk.js/Proxy/Core/AbstractRepresentationProxy';
+  export interface vtkGeometryRepresentationProxy extends vtkAbstractRepresentationProxy {
+    setVisibility(visible: boolean): boolean;
+    getVisibility(): boolean;
+  }
+
+  export default vtkGeometryRepresentationProxy;
+}
```
finetjul, line 294: it would be better to directly expose those in VTK.js

src/vtk/ObliqueRepresentationProxy/index.js
```diff
@@ -0,0 +1,121 @@
...
+    const tmp = [0, 0, 0];
+    for (let i = 0; i < points.length; i += 3) {
+      const p = points.subarray(i, i + 3);
+      if (transform.length === 9) {
+        vec3.transformMat3(tmp, p, transform);
+      } else {
+        vec3.transformMat4(tmp, p, transform);
+      }
+      [p[0], p[1], p[2]] = tmp;
+    }
```
finetjul, line 24: `vtkMatrixBuilder.apply` does that already

src/vtk/ObliqueRepresentationProxy/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+    model.cubeSource.update();
+    const out = model.cubeSource.getOutputData();
+    outPD.getPoints().setData(Float32Array.from(out.getPoints().getData()), 3);
+    outPD.getPolys().setData(Uint32Array.from(out.getPolys().getData()), 1);
+
+    // Now, transform the cube polydata points in-place
+    // using the image's indexToWorld transformation.
+    const points = outPD.getPoints().getData();
+    transformPoints(points, image.getIndexToWorld());
```
finetjul, line 55: why not set the matrix to the cubesource ?

src/vtk/ObliqueRepresentationProxy/index.js
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  model.cubeSource = vtkCubeSource.newInstance();
+  model.cubePolyData = vtkPolyData.newInstance();
```
finetjul, line 59: it is not very VTK-ish to create intermediate data objects.
this should be handled with pipelines i.e.  `m.cutter.setInputConnection(model.cubeSource.getOutputPort())`

src/vtk/ObliqueRepresentationProxy/index.js
```diff
@@ -0,0 +1,121 @@
...
+      model.cubeSource.setXLength(1);
+      model.cubeSource.setYLength(1);
+      model.cubeSource.setZLength(1);
+    }
+
+    model.cubeSource.setCenter(
+      model.cubeSource.getXLength() / 2.0,
+      model.cubeSource.getYLength() / 2.0,
+      model.cubeSource.getZLength() / 2.0
+    );
```
finetjul, line 45: call `model.cubeSource.setBounds(sext)` instead

src/vtk/ObliqueRepresentationProxy/index.js
jadh4v: Yes, this is due for refactor after I added the ImageDataOutlineFilter to vtk-js.

src/vtk/ObliqueRepresentationProxy/index.js
jadh4v: This is refactored into vtk-js here: https://github.com/Kitware/vtk-js/pull/2908
Will be removing ObliqueRepresentationProxy from Volview.

src/vtk/ObliqueRepresentationProxy/index.js
jadh4v: This is refactored into vtk-js here: https://github.com/Kitware/vtk-js/pull/2908
Will be removing ObliqueRepresentationProxy from Volview.

src/shims-vtk.d.ts
jadh4v: Typescript definitions added here: https://github.com/Kitware/vtk-js/pull/2906

src/components/VtkObliqueView.vue
```diff
@@ -0,0 +1,699 @@
...
+        const x = corners[i][0];
+        if (x > maxX) {
+          maxX = x;
+        }
+        if (x < minX) {
+          minX = x;
+        }
+      }
+
+      return [roundTo2Decimals(minX), roundTo2Decimals(maxX)];
```
floryst, line 372: What's the purpose of rounding to 2 decimal points here?

src/config.ts
```diff
@@ -63,7 +95,7 @@ export const Layouts: Record<string, Layout> = [
   {
     name: 'Axial Only',
     direction: LayoutDirection.H,
-    items: [InitViewIDs.Axial],
+    items: [InitViewIDs.Sagittal],
```
floryst, line 98: This should remain Axial.

src/config.ts
```diff
@@ -101,6 +133,20 @@ export const Layouts: Record<string, Layout> = [
       },
     ],
   },
+  {
+    name: 'Oblique Only',
+    direction: LayoutDirection.H,
+    items: [
+      {
+        direction: LayoutDirection.V,
+        // items: [InitViewIDs.ObliqueCoronal],
```
floryst, line 142: Extraneous comment

src/store/custom-events.ts
floryst, line 1: This can be replaced with [`createEventHook`](https://vueuse.org/shared/createEventHook/).

jadh4v: Ok, I will investigate this feature and replace.

src/components/VtkObliqueView.vue
jadh4v: Initially, I was trying to have some "meaningful" numbering for the oblique slices, but I don't think quantifying that is useful for non-axis-aligned slices. I am also thinking of removing the individual reset-camera buttons as they actually affect all three views. The global button on the toolbar should be sufficient.

jadh4v: The rounding was since the numbers were displayed in the overlay.

jadh4v:
@floryst I think this is ready. I have added a `MultiSliceRepresentationProxy` as a container for the 3D oblique view which will hold 3 new oblique representations along with the image outline. This way the initialization problem is gone. It works well with the `useSceneBuilder()` as I don't have to construct anything in the `setup()` of the `ObliqueThreeView` component.
The `MultiSliceRepresentationProxy` could be generalized and pushed into vtk-js in the future (maybe as `vtkCompositeRepresentationProxy`). But I am not entirely sure it has a general use case to warrant a promotion into vtk-js.

I have also removed the `custom-events` store and using `createEventHook` instead, as you suggested.

Additional fixes listed [in the PR description above](https://github.com/Kitware/VolView/pull/359#additional-fixes).

src/components/App.vue
```diff
@@ -387,6 +396,56 @@ export default defineComponent({
       }
     });
 
+    // --- ResliceCursorWidget --- //
+    // Construct the common instance of vtkResliceCursorWidget and provide it
+    // to all the child ObliqueView components.
+    const resliceCursor = ref<vtkResliceCursorWidget>(
```
floryst, line 402: Is it necessary for the reslice cursor widget to be a ref? If not, we can drop ref here.

floryst (commented):
This should be a good starting point. We can make refactors as we go. Only one minor comment relating to the resliceCursor.

src/components/App.vue
jadh4v: Yes, it doesn't have to be ref. Fixed it in a new commit.

aylward:
Looking good, but seems to still have some issues (that might not be specific to oblique views:

1) Crashes when switching to another view.  Steps to replicate: (a) load ultrasound sample. (b) switch to oblique view. (c) close view menu. (d) open view menu. (e) switch to quad view - crashes on my nvidia 3060 laptop.   Happens most often with ultrasound data, but does sometimes happen with other data.
Lost WebGL context...
image

2) Changing rendering transfer function crashes.  Steps: (a) Load brain MRA Head and Neck data, (b) switch to rendering tab, (c) slide left/right the transfer function opacity curve - crashes
Lost WebGL context...
and
image
and
image
image


A few suggestions
1) When you choose a layout using the layout menu, the menu should immediately close.  It is unlikely that people will want to select multiple layouts in a row.
2) Orientation of images in the 3 planes is NOT the same as the orientation of images in the quad view for the ultrasound data - that data is shown upside down in most subwindows.
3) Use parallel projection - not perspective projection for 3D view

This work needs much more hands-on testing.

aylward:
Looking good!  Is not crashing.  Orientation is fixed.  

For me the layout menu still doesn't auto-close when an option has been selected.

My preference is to continue to have sliders for moving between slices in each view - but I realize that this is a bit odd, since the amount you should step when you advance a "slice" in an oblique direction is not well defined.    However, I think it would be useful in practice.   Could you set the step size to the smallest of the voxel spacing values for every window and then provide sliders similar to the sliders in the quad / axial views?

If you don't have time for the slider or for auto-closing the layout menu now, you can merge as-is, and I can make an issue/feature requests instead.

Overall, this is outstanding!

aylward (approved):
LGTM!

jadh4v:
> Looking good! Is not crashing. Orientation is fixed.
> 
> For me the layout menu still doesn't auto-close when an option has been selected.
> 
> My preference is to continue to have sliders for moving between slices in each view - but I realize that this is a bit odd, since the amount you should step when you advance a "slice" in an oblique direction is not well defined. However, I think it would be useful in practice. Could you set the step size to the smallest of the voxel spacing values for every window and then provide sliders similar to the sliders in the quad / axial views?
> 
> If you don't have time for the slider or for auto-closing the layout menu now, you can merge as-is, and I can make an issue/feature requests instead.
> 
> Overall, this is outstanding!

Yes, setting the step size to the smallest spacing would be good empirical default I think. I will check if this and the menu closing can be done quickly / easily.

finetjul:
> Could you set the step size to the smallest of the voxel spacing values for every window and then provide sliders similar to the sliders in the quad / axial views?



For some reason (can't recall why), we used a slightly different approach within ResliceCursorWidget:
https://github.com/Kitware/vtk-js/blob/master/Sources/Widgets/Widgets3D/ResliceCursorWidget/behavior.js#L293-L313

The idea is that 1 step should "change" the current slice. (a tiny voxel spacing in an axis (e.g. Z) should not impact the XZ plane sliding).

I believe the code I shared is wrong because it does not apply the "spacing" on `dirProj`.

floryst:
I'll help address the e2e test failures

floryst:
I'm good to merge soon. Anything else @jadh4v?

jadh4v:
> I'm good to merge soon. Anything else @jadh4v?

Yes, let's merge. I will do another PR for adding the sliders. It will need some additional work regarding dynamic update of range and slider position (when plane rotates).

PR #361, opened 2023-07-11
Miscellaneous tool fixes

floryst:
Major changes:
- `useAnnotationTool` should create copies of defaults to avoid possibly shared object references
- fix rectangle and ruler controls on mobile


PaulHax (approved):
Gaw, mobile so broken.

Looks good.

floryst:
We'll continue to support mobile in some capacity, but not "officially".

PR #366, opened 2023-07-12
More label fixes

floryst:
Fixes the types passed into useLabels.


PaulHax (commented):
We lost some type safety with the `MakeToolDefaults extends (...args: any) => any` change earlier.  Before, ToolDefaults and label properties needed to exist in the `Ruler` or `Rectangle` type. 

Would be nice if we got a Typescript error if default label properties did not exist on tool.  Then a object of type `Label<Tool>` could not have arbitrary properties.

PaulHax (approved):
Ah, I see, thanks for the learning.

PR #367, opened 2023-07-13
Data privacy and security handling updates

floryst:
- Add more info on how we handle data privacy and security
- Add support for toggling error reporting
- Switch from GA to server side analytics to just track hits
- Remove sentry Replay and performance for now. Enabling Replay requires more testing to ensure we don't send any traces that may reveal info about a user's data
- UX: do not negate the meaning of a toggle. The "Disable error reporting" toggle in the settings now reads "Error reporting", with the toggle state indicating the on/off state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ __pycache__
 venv
 
 bundle-analysis.html
+
+# Local Netlify folder
+.netlify
--- a/index.html
+++ b/index.html
@@ -1,22 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <% if (gaId) { %>
-      <script async src="https://www.googletagmanager.com/gtag/js?id=<%= gaId %>"></script>
-      <script>
-        window.dataLayer = window.dataLayer || [];
-        function gtag(){dataLayer.push(arguments);}
-        gtag('js', new Date());
-        gtag('config', '<%= gaId %>');
-      </script>
-    <% } %>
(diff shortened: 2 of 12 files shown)


aylward:
Great start!

I like the position and the look-and-feel.

Regarding the text, I like the organization.  We should tweak the wording a bit (particularly with the development of a python server).   Also, let's turn off analytics and work with sysadmin to simply get a count of number of visitors to the weblink.
-------------
.
.
.
* Data Loading: Your image data never leaves your computer.  VolView runs independently on your computer.  Image data is loaded into your browser and is never sent to a remote server or cloud service.
* Tracking: We do not track how you are using VolView.  Your interactions with your data are not recorded or reported.
* Error Reporting: Optionally. error reports are sent to Kitware when VolView encounters an error.  These reports help us improve the stability of VolView.  These reports identify where the error occurred in our code and the current values of relevant code variables.  You cannot be identified by the information collected in these reports, and your image data is never included in these reports.

If you would like to opt-out of error reporting, you may turn it off using the following setting.  Your choice will be respected for future VolView sessions that you run on this machine.
-------------

aylward:
LGTM!

aylward (approved):

PaulHax (approved):
Works good!

I like the purple, but other buttons and toggles are different colors.  Should probably use just the Kitware blue or green as our accent someday.

floryst:
I switched the toggle to use the secondary color, similar to what the settings toggles use.

PR #368, opened 2023-07-13
Closeable dialogs

floryst:
This unifies how we size and use closeable dialogs.
- defaults for how dialogs are sized on mobile and desktop
- close button is "stickied" to the corner for scrollable content (e.g. the about info dialog)

Other notable changes:
- upgrade to vue 3.3.4. This changed some of the type inference, so the upgrade commit contains those changes.


PaulHax (approved):
Good cleanup!

Issue #363, opened 2023-07-12
Add feedback button


aylward:
Add a button at the bottom of the toolbar for sending an email to volview@kitware.com for reporting bugs and making feature requests.

tooltip for the button should be "Give Feedback"

Options for icons for this button:
![image](https://github.com/Kitware/VolView/assets/279727/914e0471-303d-43a7-9d5d-05278eb9777f)
![image](https://github.com/Kitware/VolView/assets/279727/8020f090-9c57-4c90-88a3-65d383de0c63)
![Feedback](https://github.com/Kitware/VolView/assets/279727/780f5057-9357-45a0-997f-c713a90a15de)

PR #369, opened 2023-07-13
Add user feedback button

floryst:
Addresses #363

This adds a user feedback button to the vertical toolbar.

@aylward At the moment, it's using a mailto link. We would ideally switch to a form before we merge.

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -17,7 +17,7 @@
             icon
             :rounded="0"
             class="toolbar-button"
-            href="mailto:volview@kitware.com?subject=VolView Question/Feedback"
+            href="https://volview.kitware.com/feedback/"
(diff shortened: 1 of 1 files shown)


aylward:
I think submit feedback should be above the settings and notification icons?

Perhaps the tooltip should be "Ask a question or submit feedback" - that seems to reflect the meaning of the icon being used?

Where would you like to host the page?   Could perhaps keep it on github.io so that we can easily edit it.  Here is a way to forward fields from github.io to a google doc...https://github.com/toperkin/staticFormEmails/blob/master/README.md

aylward:
LGTM!

aylward:
New link: https://volview.kitware.com/feedback/

PR #372, opened 2023-07-17
Avoids error adding PET modality image as layer

PaulHax:
Error happened in with built bundle, not with Vite dev server.

Also asks Vite to build source maps with `npm run build`


floryst:
LGTM!

PR #381, opened 2023-07-31
feat(DataBrowser): hide Sample Data if configured

### LINKED ISSUE #380: Hide Sample Data according section with config.json file
If VolView is loading data via URL parameter, integrators may want to hide the Sample Data section.

Add a entry in the config.json file to remove the Sample Data section from the Data Browser tab. 

**Sample config**
```
{
  "sampleData": {
    "hide": true
  },
  "labels": {
    "lesion": { "color": "#ff0000" },
    "tumor": { "color": "green" },
    "innocuous": {
      "color": "white"
    }
  }
}
```

PaulHax:
Kind of silly to make a new store just for this? 🤷  Simpler way?

Closes #380

### DIFF
--- a/src/io/import/processors/handleConfig.ts
+++ b/src/io/import/processors/handleConfig.ts
@@ -3,36 +3,46 @@ import { z } from 'zod';
 import { ImportHandler } from '@/src/io/import/common';
 import { useRectangleStore } from '@/src/store/tools/rectangles';
 import { useRulerStore } from '@/src/store/tools/rulers';
+import { useDataBrowserStore } from '@/src/store/data-browser';
 
-const Color = z.string();
+const color = z.string();
 
-const Label = z.object({
-  color: Color,
+const label = z.object({
+  color,
 });
 
-const RulerLabel = Label;
+const rulerLabel = label;
 
-const RectangleLabel = z.intersection(
-  Label,
+const rectangleLabel = z.intersection(
+  label,
   z.object({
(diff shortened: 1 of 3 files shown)


src/io/import/processors/handleConfig.ts
```diff
@@ -3,36 +3,46 @@ import { z } from 'zod';
...
   })
 );
 
-const Config = z.object({
-  labels: z.record(Label).or(z.null()).optional(),
-  rulerLabels: z.record(RulerLabel).or(z.null()).optional(),
-  rectangleLabels: z.record(RectangleLabel).or(z.null()).optional(),
+const sampleData = z
+  .object({
+    hide: z.boolean().optional(),
```
floryst, line 25: I would recommend making this similar in structure to the store:

```
dataBrowser: {
  hideSampleData: boolean
}
```

floryst (commented):
What do you think about adding an empty state for the sidebar if there are no panels? Something that says "No data loaded". There should be a CSS way to detect this.

I'm okay with the extra store, since it keeps with the idea that the UI should be driven by the stores.

PaulHax:
Added the "No data loaded" message.

src/io/import/processors/handleConfig.ts
PaulHax: Fixed.

floryst:
LGTM!

PR #385, opened 2023-08-07
refactor(itk-wasm): update version to 1.0.0-b.130

PaulHax:
Brings DICOM loading `.wasm`s that can grow memeory to 4GB.  So can load larger DICOM series.   

## TODO

- [x] Fix `readImageDicomFileSeries`
- [x] ~Fix emscripten stack memory leak which was worked around via patch-package~ Not seeing a heap snapshot change with 1000 iterations of  `readDicomTags`


src/io/dicom.ts
```diff
@@ -260,10 +258,12 @@ export class DICOMIO {
...
     await this.initialize();
 
-    const result = await readImageDICOMFileSeries(
-      seriesFiles.map((file) => sanitizeFile(file))
-    );
+    const inputImages = seriesFiles.map((file) => sanitizeFile(file));
+    const result = await readImageDicomFileSeries(null, {
+      inputImages,
+      singleSortedSeries: false,
+    });
```
PaulHax, line 265: @jadh4v Using the @itk-wasm/dicom package here. This call results in an error.  When I copy in a debug build of the .wasm, message is:
`CLI::ArgumentMismatch: --input-images: 1 required INPUT_BINARY_FILE:FILE missing`

Any suggestions?

PaulHax:
@floryst The patch introduced in #281 to fix a memory leak depended on `stackSave` and `stackRestore`.  These functions are no longer built-in with the prebuild .wasm files.  What do you think I should do?

floryst:
Check to see if repeated calls to readDICOMTags usage. If not, then we are good to go.

src/io/dicom.ts
jadh4v: Addressing in a PR here: https://github.com/InsightSoftwareConsortium/itk-wasm/pull/886

src/shims-itk-wasm-dicom.d.ts
```diff
@@ -0,0 +1,13 @@
+declare module '@itk-wasm/dicom' {
```
thewtex, line 1: why not use the packages's typescript declarations?

PaulHax: `@itk-wasm/dicom` package.json's `types` property points to a missing file.

thewtex: https://github.com/InsightSoftwareConsortium/itk-wasm/pull/892

floryst:
LGTM once the package.json conflicts are resolved.

PaulHax:
Conflicts resolved.  

Also need to publish and pull in new @itk-wasm/dicom package to get the 4GB memory support.  Could do that in another PR.

PaulHax:
@floryst This is good to go!

PR #394, opened 2023-08-16
Undo/redo

floryst:
- Adds core undo/redo functionality
- Adds basic undo/redo support to the tools

### DIFF
--- a/src/components/MeasurementsToolList.vue
+++ b/src/components/MeasurementsToolList.vue
@@ -4,9 +4,12 @@ import { computed } from 'vue';
 import { useCurrentImage } from '@/src/composables/useCurrentImage';
 import { AnnotationToolStore } from '@/src/store/tools/useAnnotationTool';
 import { frameOfReferenceToImageSliceAndAxis } from '@/src/utils/frameOfReference';
+import useHistoryStore from '@/src/store/history';
+import { createRemoveToolOperation } from '@/src/store/operations/tools';
+import { Store } from 'pinia';
 
 const props = defineProps<{
-  toolStore: AnnotationToolStore<ToolID>;
+  toolStore: AnnotationToolStore<ToolID> & Store;
   icon: string;
 }>();
 
@@ -33,7 +36,10 @@ const tools = computed(() => {
 });
 
 const remove = (id: ToolID) => {
-  props.toolStore.removeTool(id);
+  const imageID = currentImageID.value;
+  if (!imageID) return;
+  const op = createRemoveToolOperation(props.toolStore, id);
+  useHistoryStore().pushOperation({ datasetID: imageID }, op, true);
 };
 
 const jumpTo = (id: ToolID) => {
--- a/src/components/tools/polygon/PolygonTool.vue
+++ b/src/components/tools/polygon/PolygonTool.vue
@@ -59,7 +59,10 @@ import {
 } from '@/src/utils/frameOfReference';
 import { usePolygonStore } from '@/src/store/tools/polygons';
 import { Polygon, PolygonID } from '@/src/types/polygon';
-import { createAddToolOperation } from '@/src/store/operations/tools';
+import {
+  createAddToolOperation,
+  createRemoveToolOperation,
+} from '@/src/store/operations/tools';
 import useHistoryStore from '@/src/store/history';
 import { IHistoryOperation } from '@/src/types/history';
 import { Maybe } from '@/src/types';
@@ -221,7 +224,13 @@ export default defineComponent({
     };
 
     const deleteToolFromContextMenu = () => {
-      activeToolStore.removeTool(contextMenu.forToolID);
+      const imageID = currentImageID.value;
+      if (!imageID) return;
+      const op = createRemoveToolOperation(
+        activeToolStore,
+        contextMenu.forToolID
+      );
+      useHistoryStore().pushOperation({ datasetID: imageID }, op, true);
     };
 
     // --- tool data --- //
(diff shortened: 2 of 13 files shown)


src/components/MeasurementsToolList.vue
```diff
@@ -33,7 +36,10 @@ const tools = computed(() => {
 });
 
 const remove = (id: ToolID) => {
-  props.toolStore.removeTool(id);
+  const imageID = currentImageID.value;
+  if (!imageID) return;
```
PaulHax, line 40: No imageID, but showing remove button in toolList would be strange.  Throw Error or derive imageID from toolID so imageID is guarrentied?  Worried that defensive programming could lead to corner case runtime quirks.

src/store/operations/tools.ts
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  return {
+    apply,
+    revert,
+    isApplied,
+  };
+}
+
+export function createUpdateToolOperation<
```
PaulHax, line 65:
```suggestion
// Not used at the moment
export function createUpdateToolOperation<
```

src/components/tools/polygon/PolygonTool.vue
```diff
@@ -154,11 +169,13 @@ export default defineComponent({
...
-        placingToolID.value = activeToolStore.addTool({
-          imageID: currentImageID.value,
-          placing: true,
-        });
+      const imageID = currentImageID.value;
+      if (imageID) {
+        useHistoryStore().pushOperation(
+          { datasetID: imageID },
+          addToolOperation!
+        );
```
PaulHax, line 177: What do you think about hiding `pushOperation` and its ilk in the tool stores?

src/components/MeasurementsToolList.vue
```diff
@@ -4,9 +4,12 @@ import { computed } from 'vue';
 import { useCurrentImage } from '@/src/composables/useCurrentImage';
 import { AnnotationToolStore } from '@/src/store/tools/useAnnotationTool';
 import { frameOfReferenceToImageSliceAndAxis } from '@/src/utils/frameOfReference';
+import useHistoryStore from '@/src/store/history';
+import { createRemoveToolOperation } from '@/src/store/operations/tools';
+import { Store } from 'pinia';
 
 const props = defineProps<{
-  toolStore: AnnotationToolStore<ToolID>;
+  toolStore: AnnotationToolStore<ToolID> & Store;
```
PaulHax, line 12: In the future, should I put `& Store` on the `AnnotationToolStore` shared type?

PaulHax (approved):
Works good!  Tests even, wowy.

floryst:
I'm identifying possible shortcomings with the current architecture: hierarchical transactions. One motivating example is the polygon tool: when placing points, undo would ideally remove dropped points, but when the polygon is finalized, undo should remove the entire polygon and not individual points.

PaulHax:
Would be cool if we made the polygon resumable.  So can drop a few points, change tools/slice, switch back and close the polygon.  Can then drop placing flag headache, it becomes "closed" for polygon case.  

Avoids the "hierarchical transactions" problem for this case.

floryst:
> So can drop a few points, change tools/slice, switch back and close the polygon.

I'm not sure this is desired behavior, only because I've never observed such behavior. I've always considered editing an ephemeral state.

In any case, I think we still need some semblance of a hierarchical/transactional history mechanism, since tools may have sub-histories only relevant to their operation.

src/store/__tests__/history.spec.ts
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable no-param-reassign */
+import useHistoryStore from '@/src/store/history';
+import { expect } from 'chai';
+import { createPinia, setActivePinia } from 'pinia';
+import { it, beforeEach, describe } from 'vitest';
+
+interface Collection {
```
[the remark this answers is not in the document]
floryst: This is just an interface for a test, so I'm not too concerned about the `I` prefix.

src/store/operations/tools.ts
floryst: I intended to add support for undoing stuff like changing labels, moving rectangle points, etc. That hasn't happened yet, so I'm good to omit this function for now instead of adding a comment that we will have to later remove anyways.

src/components/MeasurementsToolList.vue
floryst: Yeah, I think that is a good idea. I could add that as an extra commit to this PR if you'd like.

src/types/history.ts
```diff
@@ -0,0 +1,11 @@
+export interface IHistoryOperation<T = void> {
+  apply(): T;
+  revert(): void;
+  isApplied(): boolean;
```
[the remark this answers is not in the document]
floryst: I left `isApplied` as just a way for other code to know if an operation has been applied or reverted, and a convenient way to perform idempotency checks in the operation apply/revert code. HistoryManager itself implicitly keeps track of this state via the undo/redo stacks.

src/components/MeasurementsToolList.vue
floryst: I like deriving the image ID from the tool ID in this case.

src/components/tools/polygon/PolygonTool.vue
floryst: It might be good to do it later down the road. I'm opting not to for now because I don't want to make the tool stores undo/redo aware just yet.

floryst:
Moving to draft. At the moment, the integration is not there, but this PR has logic that may be useful for undo/redo implementation in the future.

PR #396, opened 2023-08-22
Polygon Tool: Add + Delete Points

### LINKED ISSUE #388: Polygon Tool: Add/Delete points
After a polygon is placed:

Right-clicking on a *point* pops the context menu with options: "Delete Point" and "Delete Polygon".

Right-clicking on a *edge* pops a context menu with options: "Add Point" and "Delete Polygon".

PaulHax:
Left click polygon line to add a point.  Right click a point to show delete point context menu.

TODO

- [ ] immediately set added point to "activeState" after click.  At the moment have to move mouse (ping WidgetManager to updateSelection?) 
- [x] Hold left button to immediately drag added point
- [x] Docs

Closes #388


PaulHax:
After I drop a new Polygon handle, it is under the mouse.  But the next click starts a new polygon instead of triggering a grab on the handle.  WidgetManager only updates selection on mouse move.

 Is there a way to poke WidgetManger to run its `updateSelection()` manualy?  Craft an event for the renderWindow interacter?

floryst:
Maybe unrelated, but I wonder if picking during animation will help here. I have some pending changes to the WidgetManager to support this.

PaulHax:
If the widget behavior could trigger an animation on left click and the handle under the mouse gets selected, that would do it.  

It might make sense to have just have "widgetManager.setSelectedWidget" so widget behaviors could trigger "dragging" state syncronously right after button down?

floryst:
Would `widgetManager.grabFocus(widget)` work for your use case?

PaulHax:
> Would widgetManager.grabFocus(widget) work for your use case?

That would be perfect.  Let me try.

PaulHax:
Regarding synchronously selecting/dragging a handle:

> Would widgetManager.grabFocus(widget) work for your use case?

This was the right concept... and the right function was `publicWidgetAPI.activateHandle()`

Still have a slight behavior bugaboo: Can't immidatly drag a handle after finishing a polygon without mouse movement.   But this PR is ready for review.

src/vtk/PolygonWidget/behavior.ts
```diff
@@ -19,6 +21,25 @@ export default function widgetBehavior(publicAPI: any, model: any) {
...
+  // widgetManager.getSelections() still points to the last actor
+  // after the mouse button is released. In this widgets case the last actor is part of the LineGlyphRepresentation.
+  // So overUnselectedHandle tracks if the mouse is over a handle so we don't create a new handle
+  // when clicking on segment without immediate mouse move.
+  // A mouse move event sets overUnselectedHandle to false as we can then rely on getSelections().
+  let overUnselectedHandle = false;
+
+  // Check if mouse is over segment between handles
+  const checkOverSegment = () => {
+    // overSegment guards against clicking anywhere in view or over seasoned handle.
```
floryst, line 34: What is a "seasoned" handle?

src/vtk/PolygonWidget/behavior.ts
```diff
@@ -84,23 +106,25 @@ export default function widgetBehavior(publicAPI: any, model: any) {
...
-    );
+    return manipulator.handleEvent(callData, model._apiSpecificRenderWindow);
+  }
+
+  function updateActiveStateHandle(callData: any) {
+    const worldCoords = getWorldCoords(callData);
 
     if (
+      worldCoords &&
       worldCoords.length &&
```
floryst, line 124: Nit: this can be simplified to `worldCoords?.length`

floryst (commented):
Will this be waiting for the LineGlyphRepresentation from your vtk.js PR?

src/components/tools/polygon/PolygonWidget2D.vue
```diff
@@ -118,9 +118,12 @@ export default defineComponent({
...
       rightClickSub = widget.value.onRightClickEvent((eventData) => {
-        const coords = getCSSCoordinatesFromEvent(eventData);
-        if (coords) {
-          emit('contextmenu', coords);
+        const displayXY = getCSSCoordinatesFromEvent(eventData);
+        if (displayXY) {
+          emit('contextmenu', {
+            displayXY,
+            widgetActions: eventData.widgetActions,
+          } satisfies ContextMenuEvent);
```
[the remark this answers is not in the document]
PaulHax: More a note.  But it does error, like in this case:
```
          emit('contextmenu', {
            coords: displayXY, // error on key name
            widgetActions: eventData.widgetActions,
          } satisfies ContextMenuEvent);
```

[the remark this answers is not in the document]
floryst: In this usage, I see no difference in using `satisfies` vs `as`. I'm good to keep it as-is, since emit isn't typed to support it.

src/vtk/PolygonWidget/behavior.ts
PaulHax: I'll update comment to this:
```
    // overSegment guards against clicking anywhere in view or
    // if over handle that was created before a mouse move event.
```

Also, you can hold off on the testing.  Behavior.ts is horrably broken right now.

src/vtk/PolygonWidget/behavior.ts
PaulHax: 🙏

src/components/tools/polygon/PolygonWidget2D.vue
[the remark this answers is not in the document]
PaulHax: `as` won't error, `satisfies` will.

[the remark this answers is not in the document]
PaulHax: Ooo, au'contrar, looks like `as` does error.

src/vtk/PolygonWidget/behavior.ts
PaulHax: Found a couple more logic simplifications along `?` lines.

PaulHax:
Don't think we should wait for LineGlyphRepresentation PR to land in VTK.js.  LineGlyphRepresentation files are idential now, we can switch over to VTK.js import latter.

PR good to go... again.

floryst (approved):
LGTM!

For the future, we will look into improving the WidgetManager's picking scenarios. Additionally, when we start supporting dragging annotations, adding points might become a right click context menu action.

PR #398, opened 2023-08-25
chore(vite.config): move port config to package.json script

PaulHax:
I'm finding the new vite port and its fallback logic a little more convenent.  `npm run dev` gets me there now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "vite",
     "preview": "vite preview",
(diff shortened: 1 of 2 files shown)


floryst:
What fallback logic are you referring to? I kept to 8080 to keep with pre-vite behavior, but since this is a dev detail I'm okay with changing it. I will request that `serve` not use a different port but just be an alias to `dev`, since having two dev build commands that are on different ports is confusing.

PR #399, opened 2023-08-25
chore: bump chromedriver to 116

floryst:
This should get the tests running again.


floryst:
macOS is being finicky. I'll look to see if we can pin to v115 or something.

PR #402, opened 2023-09-01
fix(vite): generically resolve module path

floryst:
Resolves itk-wasm paths more generically, i.e. when VolView is a submodule of another project.

Depends on https://github.com/InsightSoftwareConsortium/itk-wasm/pull/923.


floryst:
Update: the windows snapshot is now at 1200x800 and the windows runner should no longer hang.

floryst:
The windows test seems flaky, but I'll use it for now. If it continues to flake, I'm going to disable it.

PR #404, opened 2023-09-05
Annotation Tool Hover

PaulHax:
Hover popup appears 250ms after cursor rests on invisible vtkWidgetRepresentation.

Current look:
image

Old look:
image

### DIFF
--- a/src/components/tools/AnnotationInfo.vue
+++ b/src/components/tools/AnnotationInfo.vue
@@ -47,7 +47,7 @@ const offset = computed(() => {
     :style="{
       left: `${info.displayXY[0] - offset.x}px`,
       top: `${info.displayXY[1] - offset.y}px`,
-      zIndex: 500,
+      zIndex: 500, // stay under context menu
     }"
     class="better-contrast"
   >
--- a/src/components/tools/AnnotationInfo.vue
+++ b/src/components/tools/AnnotationInfo.vue
@@ -1,15 +1,20 @@
 <script setup lang="ts" generic="ToolID extends string">
 /* global ToolID:readonly */
+import { computed, ref } from 'vue';
+import { useElementSize } from '@vueuse/core';
 import { AnnotationToolStore } from '@/src/store/tools/useAnnotationTool';
 import { OverlayInfo } from '@/src/composables/annotationTool';
-import { computed } from 'vue';
+
+// These seem to work ¯\_(ツ)_/¯
+const TOOLTIP_PADDING_X = 30;
+const TOOLTIP_PADDING_Y = 20;
 
 const props = defineProps<{
   info: OverlayInfo<ToolID>;
   toolStore: AnnotationToolStore<ToolID>;
 }>();
 
-const menu = computed(() => {
+const visible = computed(() => {
   return props.info.visible;
 });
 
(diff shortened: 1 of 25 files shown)


floryst:
- Maybe we should switch from a bounding box test to picking the closest tool. I cannot see the label for a polygon inside another polygon.
- Having the label be associated with the mouse position seems a little distracting. What do you think about a fixed offset from the tool's bounding box?

PaulHax:
> Maybe we should switch from a bounding box test to picking the closest tool. I cannot see the label for a polygon inside another polygon.

I'll try getting distances to each line to sort among tools the pointer is inside.  Getting fancy.  First pass I had earler was to show popover when hovering over handles or lines, same target area as context menu.  Was probably going to do that approach that for Ruler, as bounding box could be a bit much for the diagonal ruler?   Maybe we go with that hardware picker approach for all tools for now and refine our "2D picking" approach later?

> Having the label be associated with the mouse position seems a little distracting. What do you think about a fixed offset from the tool's bounding box?

Argeed. Can't escape that needy popover.  🤣  Did not do the offset from bounding box cuz if camera is zoomed in and annotation bounds is out of frame, won't see popover.  Was common situation in catheter based modality project i worked on.   How about a debouce on mouse movement before showing popover?

PaulHax:
Thanks for checking on this early.   Good points!

src/components/tools/AnnotationInfo.vue
```diff
@@ -0,0 +1,46 @@
...
+    <v-list density="compact">
+      <v-list-item>
+        <v-list-item-title>{{ label }}</v-list-item-title>
+      </v-list-item>
+    </v-list>
+  </v-menu>
+</template>
+
+<style>
+.popover-no-events > * {
```
floryst, line 42: Minor point: is the `> *` necessary? I thought setting pointer-events and touch-action to none applied to descendants as well.

src/components/tools/BoundingRectangle.vue
```diff
@@ -0,0 +1,75 @@
...
+  rectangle.value = {
+    x: x - handleRadius,
+    y: y - handleRadius,
+    width: maxX - x + handleDiameter,
+    height: maxY - y + handleDiameter,
+  };
+};
+
+const { points } = toRefs(props);
+watch([points], updateRectangle, { immediate: true, deep: true });
```
floryst, line 59: Minor: don't need to put `points` in an array.

src/components/tools/polygon/PolygonTool.vue
```diff
@@ -181,13 +189,27 @@ export default defineComponent({
 
     const currentTools = useCurrentTools(activeToolStore, viewAxis);
 
+    const { onHover, overlayInfo } = useHover<ToolID>(
+      currentTools,
+      currentSlice
+    );
+
+    const pointsToBound = computed(() => {
```
floryst, line 197: Minor: this name confused me, since it's just the visible tool points and not a bounding box.

src/vtk/PolygonWidget/state.ts
floryst, line 1: Are these changes used anywhere? I don't see where it's used.

floryst (commented):
Looks good so far! I presume you are going to update this for the rectangle and ruler?

floryst:
One other thing: we will do some user testing to see how people feel about the delay in popup. I don't have a better idea at the moment.

src/components/tools/AnnotationInfo.vue
PaulHax: Using `v-tooltip` now, so can remove.

src/components/tools/BoundingRectangle.vue
PaulHax: 🧹

src/vtk/PolygonWidget/state.ts
PaulHax: Removed these abandoned changes!

src/composables/annotationTool.ts
```diff
@@ -78,3 +82,104 @@ export const useRightClickContextMenu = (
...
+  watch(synchronousOverlayInfo, () => {
+    if (!synchronousOverlayInfo.value.visible)
+      overlayInfo.value = synchronousOverlayInfo.value;
+    else {
+      // Immediately set visible = false to hide overlay on mouse move, even if hovering true.
+      // Depends on widget sending hover events with every mouse move.
+      overlayInfo.value = { visible: false };
+      debouncedOverlayInfo({ ...synchronousOverlayInfo.value });
+    }
+  });
```
PaulHax, line 182: Not the prettiest variable time debounce code.  Makes me miss RxJS.

PaulHax:
Added hover support to Ruler and Rectangle.

Yes, hover feel is not great.  But do need something to get "Label" of annnotation ASAP.  Alternatives: 
- Could enable hover popup with toggle switch. (inspect mode?)  Maybe no deboounce if we did this.
- Inspect tool?
- Right click context menu shows label type.

floryst:
I like the right click context menu approach the most, especially since I have some half-written code snippet somewhere that allows you to switch the label of an annotation via the context menu.

floryst:
Overall, LGTM. The icon spacing for the context menu seems kind of big, which makes entries without icons (e.g. the polygon tool's "Delete Point") look out of place.

PaulHax:
Added blank icon to fix "Delete Point".

Added a "Select" tool that enables the hover popup.

Included `usePopperState`, thanks!

floryst:
LGTM!

PR #406, opened 2023-09-06
Batch actions for MeasurementToolList

PaulHax:
Based on branch tool-visibility.

image


src/components/AnnotationsModule.vue
```diff
@@ -1,42 +1,36 @@
...
 
-export default defineComponent({
-  components: {
-    MeasurementsRulerList,
-    MeasurementsToolList,
-    LabelmapList,
-    ToolControls,
+const tools = [
+  {
+    store: useRulerStore() as unknown as AnnotationToolStore<string>,
```
floryst, line 13: Is there a type incompatibility between `useRulerStore()` and `AnnotationToolStore<string>` that requires a cast to `unknown` first? If so, we should fix that.

src/components/MeasurementsToolList.vue
```diff
@@ -30,48 +41,140 @@ const tools = computed(() => {
...
+    .map((id) => tools.value.find((tool) => id === tool.id))
+    .filter(nonNullable)
+    .forEach(callback);
+
+function removeAll() {
+  forEachSelectedTool((tool) => tool.remove());
+  selected.value = [];
+}
+
+const globalHidden = ref(false);
```
floryst, line 89: This should instead be computed based on the aggregated state of the selection.
- If all selected tools are already hidden, it should be "show".
- If at least one selected too is visible, it should be "hide".

src/components/AnnotationsModule.vue
PaulHax: Just `useRulerStore()` works actually.  Removed cast.

src/components/MeasurementsToolList.vue
PaulHax: Fixed.

src/utils/index.ts
```diff
@@ -273,3 +273,7 @@ export function zodEnumFromObjKeys<K extends string>(
   const [firstKey, ...otherKeys] = Object.keys(obj) as K[];
   return z.enum([firstKey, ...otherKeys]);
 }
+
+export function nonNullable<T>(value: T): value is NonNullable<T> {
+  return value !== null && value !== undefined;
```
floryst, line 278: This is the equivalent condition for succinctness:

```suggestion
  return value != null;
```

src/components/AnnotationsModule.vue
```diff
@@ -1,42 +1,36 @@
...
+  {
+    store: useRectangleStore() as unknown as AnnotationToolStore<string>,
+    icon: 'mdi-vector-square',
   },
-});
+  {
+    store: usePolygonStore() as unknown as AnnotationToolStore<string>,
+    icon: 'mdi-pentagon-outline',
+  },
+];
```
floryst, line 25: No action for now, but we will revisit this later on, potentially making a lookup table mapping the Tools enum to the stores.

PaulHax: I like!

PR #409, opened 2023-09-06
refactor: drops @src alias

floryst:
`@src` is no different than `@/src`. Drop it to avoid having two options.


PaulHax:
Sounds good.

PR #410, opened 2023-09-07
Current image provider

floryst:
Adds a CurrentImageProvider, which allows for changing the return value of the `useCurrentImage()` composable.


PaulHax (approved):
Nice!

PR #413, opened 2023-09-08
refactor(onVTKEvent): easier event interface

floryst:
The `onVTKEvent` interface is more straightforward and familiar (e.g. the addEventListener interface) than the `useVTKCallback` interface.


PaulHax (approved):
Objectivly better.

PR #414, opened 2023-09-08
Simplify placing tools

floryst:
Architectural changes, using the ruler as an example:
- RulerTool has greatly simplified placing ruler management. RulerTool is now only concerned with using PlacingRulerWidget2D.
- RulerWidget2D no longer mixes placing logic and regular logic. Placing logic is now in PlacingRulerWidget2D.
- vtkRulerWidget state is now split into store-derived state and standalone state. The standalone state holds the state for the placing ruler and is resettable.
- (Placing)RulerWidget2D uses a common synchronized-state interface (`syncedState`), which is a reactive object derived from the widget state (regardless of if it's store-derived or standalone state).

Visual diagram of what these changes achieve:
image

### DIFF
--- a/src/components/tools/polygon/PlacingPolygonWidget2D.vue
+++ b/src/components/tools/polygon/PlacingPolygonWidget2D.vue
@@ -1,9 +1,4 @@
 <script lang="ts">
-import vtkRulerWidget, {
-  InteractionState,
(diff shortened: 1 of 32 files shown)


floryst:
Closing in favor of #421

PR #417, opened 2023-09-12
fix(CurrentImageProvider): move higher in component tree for MeasurementsToolList

PaulHax:


floryst:
I'm good to revert my changes while we work out the details for what it means across the entire app. Plus, having a different dataset per view makes other parts of the app a bit confusing.

floryst:
See: #418

PaulHax:
superseded by #418

PR #420, opened 2023-09-14
Widget component factory for tools

floryst:
This is a working sketch of what a component factory for the tools can look like. It is based on work done in #414.
- `createAnnotationToolComponent` is the entry point that creates a Tool component. It takes a 2D widget component and a 2D placing widget component.
- `createPlacingWidget2DComponent` and `createWidget2DComponent` do as their name implies. They rely on a common interface, `Widget2DComponentMeta`, which defines how the components operate.

image

There are a few cleanup points that need to be addressed before I can consider this to be ready.
- [ ] rename `<tool>/common.ts` to something more meaningful, e.g. something state-related.
- [ ] evaluate whether `createAnnotationToolComponent` should take render functions instead of components, similar to how the `Widget2DComponentMeta` takes a render function instead of passing in an SVG component.


floryst:
@PaulHax let me know what you think of this.

src/components/tools/common.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Construct a widget factory.
+   */
+  createWidgetFactory: (widgetState: WidgetState) => WidgetFactory;
+
+  /**
+   * A composable that syncs the widget factory's state to a reactive object.
+   */
+  useSyncedState: (widgetFactory: WidgetFactory) => SyncedState;
```
PaulHax, line 47: This is needed because PlacingWidget does not pull from a store and we want a normalized representation of state for Placing and NonPlacing components?  We do pass the `tool` to store backed NonPlacing, so passing duplicate info to NonPlacing components?

src/components/tools/createAnnotationToolComponent.ts
```diff
@@ -0,0 +1,131 @@
...
+  ToolStore extends AnnotationToolStore<ToolID>
+> {
+  type: Tools;
+  name: string;
+  useToolStore: () => ToolStore;
+  Widget2DComponent: Component;
+  PlacingWidget2DComponent: Component;
+}
+
+export function createAnnotationToolComponent<
```
PaulHax, line 30: Vue component factories??  Cool! (like React =) For my edifcation, why a factory instead of passing everything as props to some generic compoenent?  

The other naive approach would have been to just start shoving all the logic into Composables.  Probably would have ended up with one big `useWidget2D` and RulerWidget2D/PolygonWidget2D would be shells.  Opprotunity for overrides/special-casing per tool.

src/components/tools/common.ts
floryst: We are certainly passing some duplicate info, but the synced state captures some ephemeral widget state that the store might not capture. For example, the polygon tool currently stores a `movePoint`. I consider that to be ephemeral state that doesn't need to be stored, only propagated from the widget itself.

PaulHax (commented):
Better and better!  

I am a little worried that the details of WidgetStates are having to be delt with in so many places...  We are seeing the implications  "Simplify Placing Tools" PR, with its introduction of 2 representations of Widget State, we need the "useSyncedState."  Cool to have the abstraction, but its another manually defined Widget State interface per Widget.

Feel like instead of copying the state of a PlacingWidget, we consider "placing" to be a "unfinished" flag in the Pinia tool state, and the ToolWidget components can delete "placing" tools as need be.  `useAnnotationTool` can expose a `finishedTools` for use by MeasurmentToolList and serialization.   That way we keep Tool states all defined in Pinia stores.  (Right now, placing flag is like the special MoveHandle state, which I'm realziing we don't need.)

floryst:
These are good points. The duplication of state is a consequence of toying with bringing in vanilla vtk.js widgets with vanilla state, but we realistically don't need to care about vanilla state.

I'm still not a fan of putting unfinished widgets in the store, but I'm slowly warming up to the idea of it. As long as we only have just that one property associated with unfinished widgets, I'm good with keeping it. I'll take another crack at this in another branch.

floryst:
Closing in favor of #421

PR #421, opened 2023-09-15
Simplify tool code

floryst:
- Some code refactoring for easier usage
- useFrameOfReference encapsulates the frame of reference handling
- usePlacingAnnotationTool provides a more readable API for managing the placing tool
- added finishedTools to all annotation tools
- simplified the point visibility code to make it more obvious as to what it's doing
- removed the movePoint state from the polygon store

### DIFF
--- a/src/components/tools/polygon/PolygonTool.vue
+++ b/src/components/tools/polygon/PolygonTool.vue
@@ -32,15 +32,12 @@ import {
   watch,
 } from 'vue';
 import { storeToRefs } from 'pinia';
-import { vec3 } from 'gl-matrix';
 import { useCurrentImage } from '@/src/composables/useCurrentImage';
 import { useToolStore } from '@/src/store/tools';
 import { Tools } from '@/src/store/tools/types';
 import { getLPSAxisFromDir } from '@/src/utils/lps';
 import vtkWidgetManager from '@kitware/vtk.js/Widgets/Core/WidgetManager';
-import type { Vector3 } from '@kitware/vtk.js/types';
 import { LPSAxisDir } from '@/src/types/lps';
-import { FrameOfReference } from '@/src/utils/frameOfReference';
 import { usePolygonStore } from '@/src/store/tools/polygons';
 import { PolygonID } from '@/src/types/polygon';
 import {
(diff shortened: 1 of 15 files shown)


floryst:
I've decided against making one giant composable for the Tool and Widget components. I think it's an unnecessary abstraction to deal with for now if the only goal is code deduplication. Maybe we can revisit it once we've added more tools, as reducing the bundle size may become a driving reason. My primary goal is to ensure readability of the tool logic while keeping flexibility.

src/components/tools/polygon/PolygonTool.vue
```diff
@@ -194,3 +144,4 @@ export default defineComponent({
 </script>
 
 <style scoped src="@/src/components/styles/vtk-view.css"></style>
+@/src/composables/useFrameOfReference
```
PaulHax, line 147: ?

src/components/tools/ruler/RulerWidget2D.vue
```diff
@@ -162,36 +160,24 @@ export default defineComponent({
...
-      computed(() => widget.value?.getWidgetState().getFirstPoint())
-    );
-    const secondPointVisible = usePointVisibility(
-      computed(() => widget.value?.getWidgetState().getSecondPoint())
-    );
+    const widgetState = widgetFactory.getWidgetState();
+    onVTKEvent(widgetFactory.getWidgetState(), 'onModified', () => {
+      visibleStates.firstPoint = widgetState.getFirstPoint().getVisible();
+      visibleStates.secondPoint = widgetState.getSecondPoint().getVisible();
+    });
```
PaulHax, line 172: Rulers/Rectangles are not showing up upon deserialization.  This bring'em back:
```
    const widgetState = widgetFactory.getWidgetState();

    const visibleStates = reactive({
      firstPoint: widgetState.getFirstPoint().getVisible(),
      secondPoint: widgetState.getSecondPoint().getVisible(),
    });
```

src/composables/annotationTool.ts
```diff
@@ -78,3 +79,44 @@ export const useRightClickContextMenu = (
     }
   });
 };
+
+export const usePlacingAnnotationTool = <ToolID extends string>(
```
PaulHax, line 83: Nifty.

PaulHax (commented):
Good cleanup.  Small bug then good by me.

PaulHax:
Guess we should remove this too https://github.com/Kitware/VolView/blob/main/src/types/polygon.ts#L11

src/components/tools/polygon/PolygonTool.vue
floryst: Whoops! Funky line.

src/components/tools/ruler/RulerWidget2D.vue
floryst: Good catch! Fixed.

floryst:
Updated to address comments.

PR #425, opened 2023-09-19
Fix widgets not rendering after relayout

### LINKED ISSUE #424: Tools not interactive after layout switch
Reproduce:
1. Load prostate
2. Draw Ruler in upper left viewport
3. Switch Layout to Axial Only
4. Switch Layout to Quad View
5. Bug: Click to drag Ruler handle. But nothing happens

### More info
After droping another tool in the same viewport, the broken tool becomes interactive again.  

WidgetRepresentations are not getting rendered after layout switches.

floryst:
Fixes #424. The views weren't requesting a render after layout changes. I also created a `useWidgetVisibility` composable to encapsulate this behavior.

I also cleaned up toolId prop usage by directly using the prop and typing the toolId to be `PropType<ToolID>`.

### DIFF
--- a/src/composables/annotationTool.ts
+++ b/src/composables/annotationTool.ts
@@ -1,4 +1,4 @@
-import { Ref, UnwrapRef, computed, readonly, ref, watch } from 'vue';
+import { Ref, UnwrapRef, computed, onMounted, readonly, ref, watch } from 'vue';
 import { Vector2 } from '@kitware/vtk.js/types';
 import { useCurrentImage } from '@/src/composables/useCurrentImage';
 import { frameOfReferenceToImageSliceAndAxis } from '@/src/utils/frameOfReference';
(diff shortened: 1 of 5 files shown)


src/composables/annotationTool.ts
```diff
@@ -226,3 +229,32 @@ export const usePlacingAnnotationTool = <ToolID extends string>(
...
+  );
+
+  const viewProxy = computed(() => useViewStore().getViewProxy(viewId.value));
+
+  onMounted(() => {
+    if (!widget.value) {
+      return;
+    }
+    // hide handle visibility, but not picking visibility
+    widget.value.setHandleVisibility(true);
```
PaulHax, line 256:
```suggestion
    widget.value.setHandleVisibility(false);
```

PaulHax (commented):
Better but still not seeing interactive widgets after switching layouts to a view that was hidden before.

Switching tool types makes existing tools ineractive again.

src/composables/annotationTool.ts
floryst: Thanks for catching that!

floryst:
Please try again. I fixed an issue where the widgets were being added to a view of 0,0 size, which prevented proper rendering of pixel-scaled representations.

PaulHax:
Fixed now!  Devious bug, I was scratching my head there.

PR #426, opened 2023-09-19
Some refactors and cleanups

floryst:
Three minor refactors:
- getOrigin() in pointState should use toRaw to reduce access overhead. getOrigin() usage should be considered readonly for the most part anyways, since in-place mutation isn't expected to trigger modified().
- Creation of a point state interface for all point-based annotation tools.
- `useViewProxy<Mounted|Unmounted>` has been renamed to `onViewProxy<Mounted|Unmounted>` for more consistent naming. They also now take `MaybeRef<Maybe<ViewProxy>>` instead of requiring the existence of the view proxy.

### DIFF
--- a/src/components/tools/crop/Crop3D.vue
+++ b/src/components/tools/crop/Crop3D.vue
@@ -23,7 +23,7 @@ import {
   toRefs,
   watch,
 } from 'vue';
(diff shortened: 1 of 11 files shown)


PaulHax (approved):
Nice!  Are dragging handle's smoother now, or is that just in my head...

PR #427, opened 2023-09-20
refactor: drop tool-specific ID types

floryst:
The per-tool nominal types weren't proving useful. We now unify all tools to use the same ToolID type.


PaulHax (approved):
I guess TS does not have to be bunch of type gymnastics.  Enough of a workout in other ways.

floryst:
Certainly so. It's a good idea if we needed compile-and-IDE-time hints that we are using an improper ID, but turns out we never ran into that so far.

PR #434, opened 2023-09-24
feat(MeasurementsToolList): add selected annotation count

PaulHax:
Some annotation jobs require a specific number of annotations to be complete.  As a quality assurnace aid, add some text with selected annotation tool counts.

image


floryst:
LGTM

PR #438, opened 2023-09-26
feat(WebXR): Add initial support for WebXR viewing with VTK.js

tbirdso:
Adds initial support for viewing VolView scene in WebXR on the local device. Note that performance may vary widely across devices.

Depends on updates to VTK.js WebXR rendering structures in https://github.com/Kitware/vtk-js/pull/2924 .

Opening as a draft PR for discussion and feedback. Note that rendering performance may vary highly across platforms. We may want to explore optimization improvements for VTK.js WebXR rendering and/or disable some XR platform options in VolView for the initial merge.

Many thanks to @PaulHax and @floryst for helping me with VolView development 🚀 

IMG-8702

Suggested remaining TODOs:
- [ ] Explore removing the Looking Glass overlay. May require contributing a config flag to the Looking Glass WebXR polyfill.
- [ ] Test tethered VR rendering
- [ ] Test mobile VR/AR rendering
- [ ] Disable XR options with inadequate performance for now

### DIFF
--- a/src/components/RenderingModule.vue
+++ b/src/components/RenderingModule.vue
@@ -4,10 +4,11 @@ import { useCurrentImage } from '../composables/useCurrentImage';
 import VolumeProperties from './VolumeProperties.vue';
 import VolumeRendering from './VolumeRendering.vue';
 import VolumePresets from './VolumePresets.vue';
+import WebXRRendering from './WebXRRendering.vue';
 import LayerList from './LayerList.vue';
 
 export default defineComponent({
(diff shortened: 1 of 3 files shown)


tbirdso:
cc @PaulHax @floryst

tbirdso:
Note that we have observed an issue on Firefox where VolView appears to initially grab the orientation box render window rather than the volume rendering render window for its view proxy. Suggest testing on Chrome or Edge browsers instead.

PR #444, opened 2023-10-04
Add functionality to convert images to labelmaps

floryst:
- Improve image space comparison to check physical spacing
- Add support to convert an image to a labelmap for painting
- Update the labelmap store API
  - support the aforementioned functionality and future metadata
  - explicitly store ordering of labelmaps per image
- Update the UX around layering to instead be accessed via a dropdown menu


floryst:
I'm open to workflow suggestions surrounding labelmap conversion. Right now this does not support converting DICOM datasets to labelmaps.

src/utils/imageSpace.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Determines if two images occupy the same space.
+ *
+ * This will produce invalid results under certain scenarios:
+ * - image direction matrices are not invertible
+ * @param im1
+ * @param im2
+ */
+export function compareImageSpaces(
```
PaulHax, line 32: 🙏 Good learning for me here.

src/store/datasets-labelmaps.ts
```diff
@@ -119,23 +217,54 @@ export const useLabelmapStore = defineStore('labelmap', {
...
+      const { metadata } = labelMap;
+      // map parent id to new id
+      const parentImage = findImageID(dataIDMap[metadata.parentImage]);
+      metadata.parentImage = parentImage;
+
+      const newID = newLabelmapIDs[index];
+      labelmapIDMap[labelMap.id] = newID;
+
+      labelmapMetadata.value[newID] = metadata;
+      orderByParent.value[parentImage] ??= [];
```
PaulHax, line 240: 😍 I like this syntax.

PaulHax (commented):
Geting TypeError when:
1. Load Fetus
2. Immediately click Annotations tab

When converting image to labelmap, it disapears from DataBrowser. 🥹 No going back? 

Nice to have an icon or something indicate when a image is a Layer on another image.  

When do we do the great layer and labelmap unification? First step unifies the Labelmap stack on the Annotations tab with the Layers stack on the Rendering tab?

aylward:
Painted labels appear as very thin lines in orthogonal slices, and they should represent the resolution in the orthogonal direction.

Steps to reproduce: 
- Load Prostate data
- Draw in any slice
- Error: in the other windows only a thin line is drawn.

image

aylward:
Instead of presenting items at the labelmap level, they should be presented at the paint-color level.   Or maybe this should be under the labelmap grouping.     That is, if a color appears in a labelmap, it should appear as an item under that labelmap in the annotations tab, and it should be possible to rename, generate a 3D rendering, convert to contour, etc for that specific paint color.   The idea is that we're generating objects (annotations) and they can be rendered in different ways in the 2D and 3D views.   But things are presented at the object level, not at the labelmap level.

src/components/LabelmapList.vue
```diff
@@ -1,27 +1,36 @@
...
 </script>
+
+<template>
+  <v-radio-group v-model="targetPaintLabelmap">
+    <v-radio
+      v-for="labelmap in currentLabelmaps"
+      :key="labelmap.id"
+      :label="labelmap.name"
+      :value="labelmap.id"
+    />
```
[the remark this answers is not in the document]
floryst: Yeah we certainly can. I'll address styling after the functionality is ironed out.

floryst:
> Painted labels appear as very thin lines in orthogonal slices...

If you zoom in, those lines do correspond to the thickness of the painted slice.

> Instead of presenting items at the labelmap level, they should be presented at the paint-color level.

Good point. The next step is to expose an API that operates at the segmentation level. This is closely related to configurable default segmentation labels (paint labels) and how those affect labelmaps.

aylward:
You're right!   We need to include resolution (voxel size) info in the information box and give physical position of the cursor in the corner annotations.    Additionally, for the brush size we need to give physical units.  I was mistakenly thinking it was corresponding to 1 voxel, but it does not.   

This gets hard to grok when drawing things with a brush size of 1 in the different planes.   In the following I draw with brush size 1 as follows
- Green square box was drawn in axial and correctly shows as a square box in that isotropic plane
- Red boxes were drawn in the coronal and sagittal planes, and I expected them to produce a square box in the axial plane (reflecting one voxel being painted) - but it is approximating a in-plane circle, not drawing 1 voxel.   So, I think everything is ok, but we need to convey the physical size of the brush.   We should also have the option to turn off texture interpolation when rendering the slice data, so that voxel edges become more apparent and the data's resolution becomes appropriately represented.

floryst:
> When converting image to labelmap, it disapears from DataBrowser. 🥹 No going back?

I'll be adding a means of going from labelmap -> image.

> Nice to have an icon or something indicate when a image is a Layer on another image.

Agreed. I can add an icon indicating that an image is a layer. We can worry about indicating which images its a layer of when we need it.

> When do we do the great layer and labelmap unification? First step unifies the Labelmap stack on the Annotations tab with the Layers stack on the Rendering tab?

Layer + labelmap unification is more tricky, so we won't touch on it in this work. For one, overlaying other images is a purely rendering thing, while labelmaps are an annotation thing. Their colormap requirements are different (you can edit labelmap colormaps), and same with the opacity. Right now their data models are distinct enough that we should keep them apart. We can revisit this down the line.

floryst:
I've added the segment editor. Please try it out and let me know if there are any bugs/basic issues, e.g. the eraser isn't available yet.

PaulHax:
Segment editor is snazy. Basic stuff:

1. Should opacity be per labelmap like layers?
2. Should new labelmap copy all of last labelmap segments?
3. Saving a session.zip has error.
4. Should we call segments -> labels? Or vice versa...
5. Should the label editor UI match the segment editor UI or vice versa?

floryst:
> 1. Should opacity be per labelmap like layers?

Yes, that's a good idea to consider. I'm thinking we can even make opacity be applied per-segment, but I'm not sure how useful that is.

> 2. Should new labelmap copy all of last labelmap segments?

I like that behavior. I'll add it unless we have good reasons to not do such a thing otherewise.

> 3. Saving a session.zip has error.

Thanks, will look into it.

> 4. Should we call segments -> labels? Or vice versa...

This naming is mostly going off of Slicer, where a labelmap is a type of segmentation, and segments are just a particular label. I'm keeping with segments for now to reflect that, but if labels is more intuitive then we can call them labels.

> 5. Should the label editor UI match the segment editor UI or vice versa?

Good point, I haven't considered that. I think I'll try using the label editor UI for the segment editor UI.

PaulHax:
Coming together!

Do we want to keep doing this (for paint)?
https://github.com/Kitware/VolView/blob/main/src/components/ModulePanel.vue#L74-L78

What kind of review are you looking for right now? UI tweeks, code stuff, just before merge manual testing, ...?

floryst:
+1 to automatic module switching!

I suggest trying out the preview for a general feel on the functionality, bugs, and some code stuff only if you have the availability. I think the most relevant commits for you based on touched files would be 6f25f78d2a21e9498faddbb1dc0e0457c36c759e, 465ca26d60af28b717ea94a58c567142c8533031, 79cf60aa9f3243a710c161a0ab4b772168927ced, 8135d437f3254caf4aaebc71605db47f20502d75 + 674ddd0bf0665acd90231c1fb8e8bbd5595f3052, and 4aad7ded3e8c07fba641c469c20017d6b43dd685.

aylward:
Labelmap naming is too long and not memorable or even easily discovered:
image

- I have no easy way to map it back to the data loaded:
image

- So, instead use the name displayed in the loaded data.  For this case, "Neck 1.0 B31s"

aylward:
When choosing the paintbrush/other annotation tool, should automatically switch to the annotation tab.  The data will have been chosen already, and the rendering controls should have already been set, prior to painting in which the first step after clicking on the paintbrush tool is to select a label, radius, etc.

aylward:
"Edit Labelmap" should be "Edit Labelmap Name" - unless there will be more option to edit in the near future?
image

aylward:
What is "Convert to image" going to do?   When should a labelmap become an image instead of remaining as a labelmap?

floryst:
Updated:
- Removed Convert to Image. This functionality will be reserved for a future feature PR, unless we consider it to be necessary.
- Edit Labelmap -> Edit Name
- Annotation tab is now focused when the paintbrush is selected
- DICOM image names are now derived from the series description

aylward:
Annotation tool option are not consistently placed

For paint: Labels are below labelmap for paint while radius and opacity are above Measurements:
image

For rectangle: Labels are above Measurements:
image

For polyline, Labels are above measurements, but label colors and names are inconsistent with rectangle - label colors should be the same regardless of tool being used:
image

aylward:
Not certain how to change the color/label for a polygon tool.
Steps:
1) Create polygon in white
2) Try to change just that polygon to be a different label, etc green.  

Thought it should be possible to click on the color/label in the measurement list and choose a new label.

aylward:
1) Seems like labelmap should be an option under the paint controls - not its own / independent listing.
2) When you paint in a label/color, I think it should be listed as a measurement.  Then you can delete it,  show/hide it, and 
3) Eventually we should list statistics associated with each are measure
      - area
      - mean intensity
      - standard deviation of intensity

aylward:
How do you give a name to a ruler, square, or polyline annotation.   Much each new name have a new label?   Can I have multiple things labeled and painted as "tumor" and yet name each one independently, e.g., tumor 1, tumor big, maybe-its-a-tumor tumor...

floryst:
Let's discuss the behaviors for annotation labels, since I think there are some conflicting expectations.

floryst:
Updated:
- measurements and labelmaps are now organized into tabs. Selecting certain annotation tools will focus the tab.

I do not plan on addressing the expected behavior of labels in this PR because this is intended to only be related to labelmaps.

floryst:
Update:
- add eraser
- merged main into this branch

@aylward @finetjul how does the following naming seem to you?
- Labelmap -> SegmentGroup
- Label/Segment -> SegmentMask (runner-up suggestion: Segment)

floryst:
Updated:
- eraser is now global and available via a tool mode swatch in the paint controls
- user is prompted for a new name when a labelmap is created
- dropdown menu has been replaced in favor of a radio list of labelmaps

floryst:
@aylward @PaulHax this is ready for functional evaluation again. Ignore the naming for now.

floryst:
I'll probably merge this soon and we can iterate on it after the fact.

PR #449, opened 2023-10-08
fix(VtkTwoView): layer opacity has uniform 1 opacity transfer function 

### LINKED ISSUE #355: Opacity not linear / underestimated ?
Hi there,

I have a question about Opacity when fusing PET and CT images, the Opacity slider is wrong,

Here a comparison of the VolView and Fiji PET/CT Viewer in the publicly available Orthanc Dataset (https://demo.orthanc-server.com/app/explorer.html#study?uuid=6b9e19d9-62094390-5f9ddb01-4a191ae7-9766b715)

Opacity of 0.5

VolView : 
image

Fiji : 
image

Opacity of 1 
VolView : 
image

Fiji : 
image

As you can see the Opacity in VolView is much less intensity than in Fiji.
The Opacity =  1 example illustrate the most the issue, at Opacity = 1 for PET layer the CT should not be visible at all because PET hide it completly, that the behaviour in Fiji but in VolView the CT is still clearly visible.

The Opacity is time to time used by physician to switch between volume being displayed ( 0 =  only CT and 1 = only PET) and the opacity should be linear 

In Volview I have the impression the slider give a range of 0 to  0.5 opacity (or even less maybe 0 - 0.4)

Thanks for any information about that,

Best regards,

Salim

PaulHax:
Removes voxel intensity based opacity transfer for layer slices.  All layer voxels have opacity transfer function Y value of 1.  Layer opacity controled by slice mesh opacity.

Good CT-PT dataset for testing https://demo.orthanc-server.com/app/explorer.html#study?uuid=6b9e19d9-62094390-5f9ddb01-4a191ae7-9766b715
    

Closes #355

### DIFF
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -203,7 +203,7 @@ import BoundingRectangle from '@/src/components/tools/BoundingRectangle.vue';
 import { useToolSelectionStore } from '@/src/store/tools/toolSelection';
 import { useAnnotationToolStore } from '@/src/store/tools';
 import { doesToolFrameMatchViewAxis } from '@/src/composables/annotationTool';
-import { TypedArray } from '@kitware/vtk.js/types';
+import type { TypedArray } from '@kitware/vtk.js/types';
 import { useResizeObserver } from '../composables/useResizeObserver';
 import { useOrientationLabels } from '../composables/useOrientationLabels';
 import { getLPSAxisFromDir } from '../utils/lps';
--- a/src/components/VtkTwoView.vue
+++ b/src/components/VtkTwoView.vue
@@ -241,7 +241,6 @@ import {
   WL_HIST_BINS,
 } from '../constants';
 import { useProxyManager } from '../composables/proxyManager';
-import { getShiftedOpacityFromPreset } from '../utils/vtk-helpers';
 import { useLayersStore } from '../store/datasets-layers';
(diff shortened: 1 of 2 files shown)


salimkanoun:
Hi there, the fusion is better but i think there still a problem, from 0.8 to 1 I don't see real change in opacity.

I see the opacity seems to be defined using gaussians, points or node while I think for PET/CT is should be a simple constant for all pixel in the image which balance with the CT opacity ( changing 0 to 1 with always Opacity CT + Opacity PT = 1) 

For a layer are all voxel opacity a constant in the current code ?

PaulHax:
@salimkanoun Your right, .8 to 1 does nothing.  This PR was trying to preserve CT context with some intensity to opacity mapping.  Uniform opacity across image would be too easy =)  

But you've finaly gotten through to me.  I'll make the PET image's opacity uniform across the image, and the slider will change the opacity of the higher layer uniformly.

salimkanoun:
Sounds good, will try the new implementation, thanks !

PaulHax:
I pushed up the simpler layer blending approach.

salimkanoun:
Seems to me much better ! Thanks !

floryst:
LGTM!

floryst:
@PaulHax we can merge this once it's rebased.

PaulHax:
Rebased!

PR #451, opened 2023-10-09
fix(annotation-tools): handle radius too small with devicePixelRatio

PaulHax:
No need to scale SVG circle radius by devicePixelRatio.

Before with devicePixelRatio 1
image

Before with devicePixelRatio 2
image

After with devicePixelRatio 1
image

After with devicePixelRatio 2
image


floryst:
LGTM overall. I think we should reduce the handle size by half since the handles are now much larger than before.

PaulHax:
No strong opinion on handle size.  If reduced, they would get smaller for us devicePixelRatio 1 peeps.

floryst:
Thanks for the clarification, I did mean the visual handle size. The pickable handle size can remain the same.

PaulHax:
Reduced visual handle size to 6 CSS pixels.

image

floryst:
LGTM

PR #452, opened 2023-10-10
Keyboard shortcuts cheat-sheet

PaulHax:
Pressing `?` or clicking button on setting menu pops a modal:

image

### DIFF
--- a/src/components/KeyboardShortcuts.vue
+++ b/src/components/KeyboardShortcuts.vue
@@ -1,0 +1,32 @@
+<template>
+  <closeable-dialog v-model="keyboardStore.settingsOpen">
+    <v-card>
+      <v-card-title class="d-flex flex-row align-center"
+        >Keyboard Shortcuts</v-card-title
+      >
+      <v-table class="pa-4">
+        <thead>
+          <th class="text-left">Command</th>
+          <th class="text-left">Keybinding</th>
+        </thead>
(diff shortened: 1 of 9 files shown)


src/components/KeyboardShortcuts.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <closeable-dialog v-model="keyboardStore.settingsOpen">
+    <v-card>
+      <v-card-title class="d-flex flex-row align-center"
+        >Keyboard Shortcuts</v-card-title
```
floryst, line 5: Another prettier styling snafu here. Not sure if this is something we can configure.

floryst:
Looks good! Two minor comments:
- Using esc to close the popup would be nice
- Let's try a monospace font for the letters. I think that would look more consistent.

src/components/KeyboardShortcuts.vue
PaulHax: Fixed.  🤷 I'm getting used to it.

PaulHax:
Fixed the Escape key and added monospacing on the keybindings column.

src/components/KeyboardShortcuts.vue
floryst: It might be fixed in a newer version of prettier, or there's some vue-prettier plugin we need.

PR #466, opened 2023-10-22
Label loading fix

PaulHax:
Removes favoring of existing labels in useAnnotationTool label deserialization.
Favoring existing labels broke deserialization of edited labels.

But, so config.JSON updates existing tools, add another pass to
importDataSources so config.JSON are applied after session.zip has been applied.

fixes https://github.com/Kitware/VolView/issues/454


tests/specs/session-zip.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+        watcher.close();
+        resolve();
+      }
+    });
+  });
+}
+
+describe('VolView', () => {
+  beforeEach(async () => {
+    // cleanup temp directory
```
floryst, line 42: Is this a todo?

floryst (commented):
Overall looks good!

tests/specs/session-zip.e2e.ts
PaulHax: Ooh, zombie code.  Went with the `cleanuptotal` service instead.

PaulHax:
CI passed! Quick, merge before CI thinks better of it.

But sersiouly, this PR makes somewhat pointless changes to to `pipeline.ts` so `PipelineResults` are sort of like [Eather](https://gcanti.github.io/fp-ts/modules/Either.ts.html).  I had a bug where bad Remote URLs were not getting returned as `PipelineResults` with errors.  Was thinking that was going to help me deal with nested PipelineResults where a pipeline executes on a array of PipelineResults, throwing if passed a PipelineResult with an error.  But too much Typescript for me and chickened out by not calling another `Pipeline` and just assigning priorirty in `importDataSources.ts`: https://github.com/Kitware/VolView/pull/466/files#diff-d81d57fafe24c4ee052b91ee8be768e4d1854ceecb2c6e7a9f76f1921005a202R85-R95

floryst:
The pipeline results were intended to have an Either-like aspect to them, with at-runtime distinction via the `.ok` property. The nested piplines are definitely the hardest part to deal with when it comes to specialized loading requirements. I'll look through your changes and see if there's anything we can simplify.

src/io/import/importDataSources.ts
```diff
@@ -58,34 +66,89 @@ const unhandledResource: ImportHandler = () => {
...
+  ];
+  const prioritizer = new Pipeline<
+    DataSource,
+    PriorityResult | ImportResult,
+    typeof importContext
+  >(prioritizes);
+
+  const priorityOrImportResult = await Promise.all(
+    dataSources.map((r) => prioritizer.execute(r, importContext))
+  );
```
floryst, line 79: I think a simpler approach would be to add a `loadPriority?: number` field to `ImportResult`, then sort on the priority (defaulting to 0 if the key doesn't exist). The priority conversion pipeline + buckets can then be scrapped for a single sort call (or bucket approach). I think it'll be easier to understand and reason about that way.

floryst (commented):
Splitting loading into a two-stage process (parse into loadable data sources, then load) has one minor downfall of having to wait for all resources to be ready prior to being loaded. This does make having streaming support all the more pertinent, so this two-stage approach is fine for now.

I wonder if we could have just gathered the config.json file instead of loading it and then apply it after the pipeline has run, similar to how we do it for the dicom data sources.

src/io/import/importDataSources.ts
PaulHax: Yeah, what a mess.  But better now.

Ran with your "do it like dicom sources" idea.  `handleConfig` just calls `done({ config ])`.  And the pipeline following code calls `applyConfig` on all resulting `ImportResults` with configs.

PaulHax:
Ready for review!  Disabled the new e2e test, really flakey.

aylward:
Might not be due to your changes, but sometimes the rectangle annotation is filled-in, and other times it is not. 

image

aylward:
Also, might not be your code, but the color of "artifact" (gray) doesn't match the color shown when it is drawn (yellow)

image

aylward:
The icon for rectangle includes control points...the icon for polyline does not

image

PaulHax:
@aylward Thanks for checking it out.  I just removed the fillColor property from the default rectangle labels.

Nobody liked the deformed polygon icon we had at first: 
https://pictogrammers.com/library/mdi/icon/vector-polygon/

Maybe we swap out rectangle for this: 
https://pictogrammers.com/library/mdi/icon/square-outline/

Menu of easly swapped icons here:
https://pictogrammers.com/library/mdi/

floryst:
> Maybe we swap out rectangle for this:
> https://pictogrammers.com/library/mdi/icon/square-outline/

+1 to using square-outline instead of the current one with control points.

aylward:
My concern is that there will be two similar squares next to one another -
one for the checkbox and one indicating the square shape.

Can we give the shapes a distinct background or better distinguish them
from the selection boxes?

I would also rather find a polyline with control boxes shown.  I'll
search...

aylward:
How about
https://icon-icons.com/icon/polygon-vector/208958

image

src/io/import/importDataSources.ts
```diff
@@ -58,72 +35,138 @@ const unhandledResource: ImportHandler = () => {
...
+      };
+    }
+  })();
+
+  return [
+    ...results,
+    dicomResult,
+    configResult,
+    // remove ok results that have no result data
+  ].filter((result) => !result.ok || (result.ok && result.data.length));
```
floryst, line 142: Simplification: `!result.ok || result.data.length`, since typescript will infer `result.ok` in the second branch.

src/io/import/importDataSources.ts
```diff
@@ -58,72 +35,138 @@ const unhandledResource: ImportHandler = () => {
...
       sources: importContext.dicomDataSources,
     },
   };
-  const dicomResult: PipelineResult<DataSource, ImportResult> = {
-    ok: true,
-    data: [],
-    errors: [],
-  };
+  const dicomResult = await (async () => {
+    try {
```
floryst, line 102: Let's split out these DICOM loading and config loading IIFEs into separate functions in this file for readability.

floryst (commented):
Just 2 comments, but overall LGTM.

Let's not deal with the icons in this PR. We can continue the discussion in a separate issue.

floryst:
LGTM!

PR #474, opened 2023-10-27
fix(ruler): avoid error when manipulator returns no worldCoords

PaulHax:
Seems that Paint is hurting too...

### DIFF
--- a/src/vtk/PaintWidget/behavior.ts
+++ b/src/vtk/PaintWidget/behavior.ts
@@ -1,4 +1,5 @@
 import macro from '@kitware/vtk.js/macro';
+import { computeWorldCoords } from '@/src/vtk/ToolWidgetUtils/utils';
 
 export function shouldIgnoreEvent(e: any) {
   return e.altKey || e.controlKey || e.shiftKey;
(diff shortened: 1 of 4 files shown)


floryst:
This looks to be an oversight on how we use worldCoords after the vtk.js@v28 upgrade.

> Seems that Paint is hurting too...

It sounds like all of the widgets need to be checked for their usage of `handleEvent` and worldCoords handling. If paint (and others) are having issues, please also add those fixes to this PR.

src/vtk/PaintWidget/behavior.ts
```diff
@@ -45,12 +44,10 @@ export default function widgetBehavior(publicAPI: any, model: any) {
...
     }
 
-    const { worldCoords } = model.manipulator.handleEvent(
-      eventData,
-      model._apiSpecificRenderWindow
-    )?.worldCoords;
+    const worldCoords = getWorldCoords(eventData);
 
     if (!worldCoords) {
+      console.warn('Event cannot be converted to world coordinates');
```
floryst, line 50: Redundant console.warn

src/vtk/ToolWidgetUtils/utils.ts
```diff
@@ -36,3 +36,19 @@ export function watchState(
...
+  if (!manipulator) {
+    console.error('No manipulator');
+    return undefined;
+  }
+  const { worldCoords } = manipulator.handleEvent(
+    event,
+    model._apiSpecificRenderWindow
+  );
+  if (!worldCoords)
+    console.warn('Event cannot be converted to world coordinates');
```
floryst, line 52: I think our current manipulators shouldn't trigger this unless there's an error case, so this is a good add. If we start getting them legitimately, then we can remove it.

floryst (commented):
Overall LGTM

floryst:
LGTM!

PR #476, opened 2023-10-31
fix(layers): fix initialization bugs when less than all views mounted

PaulHax:
If configured for axial only view, then add a layer, layer properties GUI would just try to get first view config (Sagittal?), which did not exist.

If a layer was created, without the view mounted, the colorBy layer config would not be set, and the layer would be invisible.

### DIFF
--- a/src/store/view-configs/layers.ts
+++ b/src/store/view-configs/layers.ts
@@ -1,4 +1,3 @@
-import vtkImageData from '@kitware/vtk.js/Common/DataModel/ImageData';
 import vtkPiecewiseFunctionProxy from '@kitware/vtk.js/Proxy/Core/PiecewiseFunctionProxy';
 import {
   getColorFunctionRangeFromPreset,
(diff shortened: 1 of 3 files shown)


src/store/view-configs/layers.ts
```diff
@@ -119,18 +118,8 @@ export const useLayerColoringStore = defineStore('layerColoring', () => {
...
-    image: vtkImageData
-  ) => {
-    const scalars = image.getPointData().getScalars();
-
-    updateColorBy(viewID, dataID, {
-      arrayName: scalars.getName() + dataID,
-      location: 'pointData',
-    });
-    setColorPreset(viewID, dataID, getPreset(dataID));
+  const defaultColorPreset = (viewID: string, layerID: LayerID) => {
```
floryst, line 121: I think this would be better named as `resetToDefault`, or at least have the verb `reset` in the name.

PaulHax: Yes, changed to `resetColorPreset`

floryst:
LGTM!

Issue #475, opened 2023-10-30
Trying to load a large image series can freeze or crash the computer


fedorov:
I thought this should not be possible, but it happens. On my Mac M2 mini, loading a large series resulted in a freeze, and @pieper reported crash. I will follow up with the URL for the sample series via direct communication on Slack to @floryst.

PR #480, opened 2023-11-04
fix(dataset-dicom): throw error if passed file but no volume results

### LINKED ISSUE #478: No image or no error on import of this DICOM file
When loading this DICOM file
[3.000000-NA-78236.zip](https://github.com/Kitware/VolView/files/13254101/3.000000-NA-78236.zip)

No image show up in data browser and green "Dataset loaded" toast showed up.  Expected seeing a loaded image or a error notification.

2 tasks:
- Show error if loading DICOM file, but VolView loads nothing
- See if we can load this file easly

### Approach

File does not load because `dicomIO.categorizeFiles` returns empty list.

PaulHax:
Closes #478

### DIFF
--- a/src/io/import/processors/importSingleFile.ts
+++ b/src/io/import/processors/importSingleFile.ts
@@ -6,6 +6,8 @@ import { useModelStore } from '@/src/store/datasets-models';
 import { FILE_READERS } from '@/src/io';
 import { ImportHandler } from '@/src/io/import/common';
 import { DataSourceWithFile } from '@/src/io/import/dataSource';
+import { useDatasetStore } from '@/src/store/datasets';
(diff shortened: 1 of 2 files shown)


aylward:
I tried dragging and dropping a .stl file, and it doesn't crash.  It reports data loaded and indicates success, but it doesn't report an error and no volume appears.

I'm using a .stl file from the Challenging Data directory.

aylward:
@floryst - can you try this with the URL provided in https://github.com/Kitware/VolView/issues/475

PaulHax:
Added a warning if loading a file that results in a PolyData, like .stl, and have not loaded an image yet.

floryst:
LGTM!

PR #483, opened 2023-11-05
fix(WindowLevelTool): Do not reset the auto/preset selection

### LINKED ISSUE #481: Auto W/L values are ignored when using mouse to then adjust W/L
How to replication

1) Select an Auto W/L setting other than full range
2) Press and hold to adjust W/L manually

Result
- W/L jumps to full range

Expected result
- W/L manual movements should adjust the current values from Auto W/L chosen, not the full range / old values.

Using 
volview-dev.kitware.app

bnmajor:
Fixes #481


bnmajor:
@aylward

floryst:
LGTM!

PR #489, opened 2023-11-07
Generate notification error when up bad URL in remoteManfest.json files

PaulHax:
* Add test expecting error notification if URL in remoteManifest.json is not a URL
* If there is a bad URL in remoteManifest.json, the datasource is passed down the Pipeline. Changed `importDataSources` `handleConfig` to pass through a JSON file that has no known "config" keys.  Then `unhandledResource`  will generate an error.


floryst:
LGTM!

PR #494, opened 2023-11-08
refactor: rename labelmaps to segment groups

floryst:
@aylward @PaulHax this goes into #444. It renames labelmap -> segment group and label -> segment (mask).

### DIFF
--- a/src/store/segmentGroups.ts
+++ b/src/store/segmentGroups.ts
@@ -8,11 +8,15 @@ import { useIdStore } from '@/src/store/id';
 import { onImageDeleted } from '@/src/composables/onImageDeleted';
 import { normalizeForStore, removeFromArray } from '@/src/utils';
 import { compareImageSpaces } from '@/src/utils/imageSpace';
-import { LabelMapSegment } from '@/src/types/labelmap';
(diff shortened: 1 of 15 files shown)


src/store/segmentGroups.ts
```diff
@@ -8,11 +8,15 @@ import { useIdStore } from '@/src/store/id';
 import { onImageDeleted } from '@/src/composables/onImageDeleted';
 import { normalizeForStore, removeFromArray } from '@/src/utils';
 import { compareImageSpaces } from '@/src/utils/imageSpace';
-import { LabelMapSegment } from '@/src/types/labelmap';
-import { DEFAULT_LABELMAP_SEGMENTS } from '@/src/config';
+import { SegmentMask } from '@/src/types/segment';
```
PaulHax, line 11: When should we prefer camelCase over kebab-case for file names?

PaulHax (approved):

src/store/segmentGroups.ts
floryst: I've been switching to use camelCase for file names, so I'd recommend doing that. It is mostly just a developer cosmetic, so I'm not looking to change all at once, and I'm not going to consider it strongly in a review.

Issue #365, opened 2023-07-12
No feedback while loading data or changing layouts


aylward:
Note also that it seems particularly hard to use right now (on my desktop, in the latest chrome). There is no feedback on how to load the sample data sets (click? double click?) and when you do click (at least for me) it freezes for 10-20 seconds with no indication of what it is doing. At least that makes some sense because it is reading data across the network. But when I switch layouts it also freezes for 10-20 seconds before actually changing the layout (and makes other chrome windows act flaky at that point too). I don't remember this happening in the past.

aylward:
Two issues
1) UX 
2) Performance

PR #497, opened 2023-11-09
Welcome page loading indicator

### LINKED ISSUE #495: Show "Loading..." on welcome screen when loading sample, URL, or local data
Will correctly inform user that loading is still happening after thumbnail appears and in other contexts.

floryst:
- Refactors out the welcome page to a separate component
- Adds a loading indicator to the welcome page

Closes #495 and partially addresses #365.

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -128,48 +128,26 @@
             </div>
             <div class="d-flex flex-column flex-grow-1">
               <layout-grid v-show="hasData" :layout="currentLayout" />
-              <v-row
-                v-show="!hasData"
-                no-gutters
-                align="center"
-                class="clickable bg-grey-darken-3"
+              <welcome-page
+                v-if="!hasData"
+                class="clickable"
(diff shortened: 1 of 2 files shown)


aylward:
Nice!

Works great for local data.

Can we also have the Loading message appear when loading data via URL, such as
https://deploy-preview-497--volview-dev.netlify.app/?urls=[https://data.kitware.com/api/v1/file/6347145311dab81428208e21/download]

s

aylward:
Maybe also while loading sample data?

Admittedly it is perhaps redundant with the "Loading data" popup, but redundancy is sometimes ok :)

floryst:
Updated:
- show loading on URL datasets

My preference is to not show the loading data screen for sample datasets since we already have an indicator within proximity to the user click action. It's not a strong preference, but one that I think makes sense.

aylward (approved):
LGTM

PaulHax (approved):
Slick!

PR #501, opened 2023-11-10
Offscreen views

floryst:
Views are rendered offscreen and kept persistent across layout changes. This should speed up layout changes and hopefully reduce potential webgl context losses from (theoretical) rapid unmount/mount of the rendering container.

Instead of having interaction and rendering occur on a single canvas, rendering is now performed on a dedicated offscreen canvas. Pixels are copied over from the rendering canvas to the interaction canvas, and interaction events are handled appropriately.

@aylward please test this out to see if it helps with context loss issues.


aylward:
Still happens - but maybe less often...

image

sedghi:
The canvas loss can stem from various root issues. Here's what we found on the OHIF side:

https://github.com/OHIF/Viewers/issues/3207

PR #502, opened 2023-11-10
Miscellaneous fixes

floryst:
- Allow selection of text in the message list
- completely disconnect the ResizeObserver on unmount
- Log out global errors and use a fallback message for error reporting

### DIFF
--- a/src/composables/useGlobalErrorHook.ts
+++ b/src/composables/useGlobalErrorHook.ts
@@ -5,7 +5,11 @@ export function useGlobalErrorHook() {
   const messageStore = useMessageStore();
 
   const onError = (event: ErrorEvent) => {
-    messageStore.addError('Application error (click for details)', event.error);
+    console.error(event);
+    const details = event.error
+      ? event.error
+      : { details: event.message ?? 'Unknown error' };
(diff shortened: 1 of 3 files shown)


src/composables/useGlobalErrorHook.ts
```diff
@@ -5,7 +5,11 @@ export function useGlobalErrorHook() {
   const messageStore = useMessageStore();
 
   const onError = (event: ErrorEvent) => {
-    messageStore.addError('Application error (click for details)', event.error);
+    console.error(event);
```
PaulHax, line 8: Good!  Thinking we should send more errors to Sentry via:
`Sentry.captureException` https://docs.sentry.io/platforms/javascript/usage/

or 
```
Sentry.init({
  dsn: 'https://your-sentry-server-dsn',
  integrations: [
    new CaptureConsole({
      levels: ['error']
    })
  ],
...
```

PaulHax (approved):
Objectively better

src/composables/useGlobalErrorHook.ts
floryst: I like this idea! I'll cover this in a later PR.

PR #503, opened 2023-11-13
fix(mimeTypes): handle windows zip mime type

### LINKED ISSUE #500: BUG: "Open files" fails to load DICOM ZIP archives
## Steps to reproduce:
1. View sample data URLs: https://github.com/Kitware/VolView/blob/main/src/config.ts#L178
2. Download `MRA-Head_and_Neck.zip`: https://data.kitware.com/api/v1/item/6352a2b311dab8142820a33b/download
3. Visit the VolView preview: https://deploy-preview-438--volview-dev.netlify.app/
4. Click "Open files" and navigate to the ZIP download

Also reproducible with other sample DICOM archive downloads from `config.ts`.

## Expected behavior: 

DICOM data loads and the MR head volume is visible in VolView

## Observed behavior: 
image

Console error:
```
Could not read the input DICOM file
o.printErr @ read-dicom-tags.umd.js:1
```

Workaround: I can drag-and-drop the zip archive on the VolView window to load. Loading by drag-and-drop succeeds, by "Open files" dialog does not.

## Versions

Latest: https://volview.kitware.app/

## Platforms

Windows 10, 11
Chrome, Firefox, Edge browsers

cc @floryst

floryst:
Fixes #500

Windows uses the non-standard `application/x-zip-compressed` mime type when opening zip files via the file picker. Other platforms (and even drag-and-drop on Windows) uses `application/zip`.

This change adds `application/x-zip-compressed` as a supported mime type.

### DIFF
--- a/src/io/mimeTypes.ts
+++ b/src/io/mimeTypes.ts
@@ -67,4 +67,8 @@ export const MIME_TYPES = new Set(Object.values(FILE_EXT_TO_MIME));
 /**
  * Supported archives
  */
-export const ARCHIVE_FILE_TYPES = new Set(['application/zip']);
+export const ARCHIVE_FILE_TYPES = new Set([
+  FILE_EXT_TO_MIME.zip,
+  // Windows uploads files with this non-standard mime type
+  'application/x-zip-compressed',
+]);


tbirdso (approved):
Much better! This resolves the loading issue for me.

PR #507, opened 2023-11-15
Paint tool on startup fixes

PaulHax:
Small paint tool fixes:
- If session.volview.zip is loaded with the paint tool initially selected, clicking on the image did not paint.
- If paint tool is activated when mouse is within a view, the stencil shape would not be correct.

### DIFF
--- a/src/store/tools/paint.ts
+++ b/src/store/tools/paint.ts
@@ -191,13 +191,14 @@ export const usePaintToolStore = defineStore('paint', () => {
   ) {
     const { paint } = manifest.tools;
     setBrushSize.call(this, paint.brushSize);
-    setActiveSegment.call(this, paint.activeSegment);
     setLabelmapOpacity.call(this, paint.labelmapOpacity);
     isActive.value = manifest.tools.current === Tools.Paint;
 
     if (paint.activeSegmentGroupID !== null) {
       activeSegmentGroupID.value =
         segmentGroupIDMap[paint.activeSegmentGroupID];
+      setActiveLabelmap(activeSegmentGroupID.value);
+      setActiveSegment.call(this, paint.activeSegment);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/store/tools/paint.ts
```diff
@@ -80,18 +80,6 @@ export const usePaintToolStore = defineStore('paint', () => {
...
-        throw new Error('Cannot set active segment without a labelmap');
-
-      const segmentGroupStore = useSegmentGroupStore();
-      const { segments } =
-        segmentGroupStore.metadataByID[activeSegmentGroupID.value];
-
-      if (!(segValue in segments.byValue))
-        throw new Error('Segment is not available for the active labelmap');
-    }
-
```
floryst, line 94: Can you explain how this block of code affected the paint tool not loading properly through a session file?

floryst:
Overall LGTM! My quick functional test didn't reveal any issues. Just one clarifying question.

src/store/tools/paint.ts
PaulHax: Yes.  `usePaintStore.deserialize` calls `setActiveSegment` before `activeSegmentGroupID` is set.  Later, activeSegmentGroupID will be set by `useToolStore.setCurrentTool` -> `usePaintStore.activateTool` -> `setActiveLabelmapFromImage`

floryst: I think then that we can solve this by calling `setActiveLabelmap(segmentGroupIDMap[paint.activeSegmentGroupID])` before `setActiveSegment` in `deserialize`. I think keeping the code in `setActiveSegment` is good to ensure we are getting a valid segment value. Thoughts?

PaulHax: Would be nice to keep the checks.   I rebased a fix.  Turns out segmentIDMap was not getting passed down to `paint.deserialize`.  Also, can't `setActiveSegment` without a segmentGroup, so only calling `setActiveSegment` if one is deserialized.

PR #513, opened 2023-11-22
fix(remotemanifest): support relative URIs without names

PaulHax:
If a resource had no `name` prop, creating a `new URL` to parse a pathname on a relative URI would error.


floryst:
LGTM!

Issue #519, opened 2023-11-27
Side panel should be collapsible and/or resizeable


timthirion:
It would be nice in a variety of circumstances to be able to collapse the side panel (Data, Annotations, Rendering). If the contained elements don't resize well, then a show/hide toggle should be sufficient.

PR #520, opened 2023-11-27
fix(sidebar): show toggle sidebar button by default

jadh4v:
Addresses: https://github.com/Kitware/VolView/issues/519

Show the "toggle sidebar" button by default on all platforms. Earlier it was only enabled for mobile devices. Showing the button on all platforms will enable the user to explicitly show/hide the sidebar as desired.


jadh4v:
@floryst

floryst:
Looks like you need to run `npm run prettify` to address the linter error.

PR #523, opened 2023-11-29
fix(CrosshairsWidget2D): avoid error when changing layout

PaulHax:
Could not create widget if View Proxy was not yet created.

Reproduce steps:
1. Activate crosshairs
2. Change view layout


floryst:
LGTM!

PR #524, opened 2023-11-29
Migrate session.volview.zip state files from 2.1.0 to 3.0.0

PaulHax:
Hacks up old manifest.JSONs to load with new schema.

Adds test for 2.1.0 session.volview.zip with labelmap.


src/io/import/processors/restoreStateFile.ts
```diff
@@ -26,14 +26,131 @@ import {
...
+  '14': {
+    value: 14,
+    name: 'Segment 14',
+    color: [0, 0, 153, 255],
+  },
+};
+
+const OLD_MANIFEST_VERSIONS = ['2.1.0', '1.1.0', '1.0.0', '0.5.0'];
+
+const migrateManifest = (manifestString: string) => {
```
floryst, line 120: It looks like that migrateManifest could be used generally for upgrading older manifests. Maybe we can have OLD_MANIFEST_VERSIONS be a map to a set of transformation functions instead?

```
UPGRADE_TRANSFORMS = {
  '2.1.0': [upgradeLabelmapsSince210],
  '1.1.0': [upgradeLabelmapsSince210],
...

const migrateManifest = (s: string) => {
  return transform(s, UPGRADE_TRANSFORMS[version] ?? [])
}
```

floryst (commented):
This is a great addition! Awesome idea. I think tweaking how we tackle upgrading now will make it easier for us down the road.

src/io/import/processors/restoreStateFile.ts
PaulHax: I like it!  Pushed up a commit showing a pattern sort of like that.  

Feel like its squarly in the middle of the curve:
https://knowyourmeme.com/memes/iq-bell-curve-midwit/photos
But it was fun.

src/utils/index.ts
floryst, line 1: There is a `utils/pipe.ts` file. Let's try to reconcile these two pipe implementations. The existing one returns a function that can be used for piping, which is different than this one.

Since this one has nicer typing, my suggestion would be to move this code to `pipe.ts` and update the old usages of `pipe` to match this new syntax.

floryst:
LGTM!

PR #535, opened 2023-12-21
Label name uniqueness tweeks

### LINKED ISSUE #477: Losing labels with matching names on serialization
### Steps:
1. Create 2 labels named "foo"
2. Save a session.volview.zip
3. Load volview.zip file
4. Bug: Only one label with name "foo" apears in label pallet.  Expected 2 labels named "foo".

### Approach

Support serialization of labels with same name.  Later, if we want to enforce unqiue label names, do that at GUI form level.  

Related: https://github.com/Kitware/VolView/issues/454

PaulHax:
Warns if user is editing label name and it maches existing label.  "Done" is disabled if name already exists.

image

Also
* New labels are created with unique names suffixed with a number.
* New labels loop through the confgi.ts:TOOL_COLORS.
* Max width on label editor dialog.

Closes #477

### DIFF
--- a/src/components/LabelControls.vue
+++ b/src/components/LabelControls.vue
@@ -105,7 +105,7 @@ function deleteEditingLabel() {
     </v-container>
   </v-card>
 
-  <isolated-dialog v-model="editDialog">
+  <isolated-dialog v-model="editDialog" max-width="800px">
     <ToolLabelEditor
       v-if="editingLabelID"
       v-model:name="editState.labelName"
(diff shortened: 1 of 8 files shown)


floryst:
I think it would be a good idea to enforce name uniqueness and not let the user create labels with duplicate names. Have we encountered a use-case for different labels with the same name?

PaulHax:
No known use case for duplicate names.  I'll disable "Done" if name is used.

floryst (approved):

floryst:
Looks like there are issues with code formatting.

PR #539, opened 2023-12-31
Export segment group GUI

PaulHax:
Add a save button for each segment group
image

Clicking pops a dialog to select export format:
image

Saving downloads the segment group to the browsers download folder.


floryst:
Nice addition! I would suggest having a dropdown selector for the file format instead of a radio.

src/components/SaveSegmentGroupDialog.vue
```diff
@@ -0,0 +1,98 @@
...
+  if (fileName.value.trim().length === 0) {
+    return;
+  }
+  try {
+    saving.value = true;
+    const image = segmentGroupStore.dataIndex[props.id];
+    const serialized = await writeImage(fileFormat.value, image);
+    saveAs(new Blob([serialized]), `${fileName.value}.${fileFormat.value}`);
+    props.close();
+  } finally {
```
floryst, line 81: If an error while saving occurs, does the UI behave/notify properly?

src/components/SaveSegmentGroupDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+  'dcm',
+  'hdf5',
+  'tif',
+  'mha',
+  'vtk',
+  'iwi.cbor',
+];
+
+const props = defineProps<{
+  close: () => void;
```
floryst, line 60: Any reason for passing close as a prop vs emitting a close event? Just to be clear, this is fine as-is, but vue tends to prefer emitting events.

PaulHax: Good tip.  React instincts. I changed it to an event.

PR #548, opened 2024-02-06
Support overriding the current image

floryst:
The `CurrentImageProvider` component can now override the image returned from `useCurrentImage()` for a given subtree. The fallback behavior is to use the current global selected image.

This revives #410 with the fallback behavior.

### DIFF
--- a/src/components/CurrentImageProvider.vue
+++ b/src/components/CurrentImageProvider.vue
@@ -1,0 +1,19 @@
+<script setup lang="ts">
+import { Maybe } from '@/src/types';
+import { provide, toRefs } from 'vue';
+import { CurrentImageInjectionKey } from '@/src/composables/useCurrentImage';
+
+const props = defineProps<{
+  imageId: Maybe<string>;
+}>();
+
+const { imageId: imageID } = toRefs(props);
+
+provide(CurrentImageInjectionKey, {
+  imageID,
+});
+</script>
+
+<template>
+  <slot />
+</template>
--- a/src/components/VtkObliqueView.vue
+++ b/src/components/VtkObliqueView.vue
@@ -219,7 +219,7 @@ export default defineComponent({
       get: () => windowingStore.getConfig(viewID.value, curImageID.value),
       set: (newValue) => {
         const imageID = curImageID.value;
-        if (imageID !== null && newValue != null) {
+        if (imageID != null && newValue != null) {
           windowingStore.updateConfig(viewID.value, imageID, newValue);
         }
       },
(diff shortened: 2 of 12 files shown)


lchauvin:
Hello,
I'm interested in using this feature, but I don't understand how it works ? How do I specify which images should be loaded by which viewer ?

Thank you.

floryst:
You can use the [CurrentImageProvider](https://github.com/Kitware/VolView/blob/main/src/components/CurrentImageProvider.vue) as outlined below. Note that this is only most relevant for making custom changes to the app, and it does make it possible for us to support viewing multiple datasets at the same time in the future.

```
<current-image-provider :image-id="myImageId">
  <vtk-slice-view ...>
</current-image-provider>
```

lchauvin:
Actually, that's what I'm trying to do (view multiple dataset). Given that the current-image-provider already exists, what is missing to be able to view multiple datasets ?

PaulHax:
You could write some code (pinia store?) "view id" gives you a `imageId` (which you feed to CurrentImageProvider of course).  This repo does not have anything like that yet.

lchauvin:
I'm gonna look at this soon, because it's something I would need :)

lchauvin:
Hello,

I'm working on specifying an image for each view, so we could display multiple images at once (like to compare images, etc...).
However, I'm wondering what UI would you prefer ? Like a list on top of each view (like Slicer), or drag'n'drop images from the list to the view ? Or something else ?

PaulHax:
I want it all. =)  

Methinks every user action does need some visual space holding button/select-dropdown/whatever so the user knows the action is possible.  Buzzword -> discoverablity.  Then we add acclerators like drag+drop and keyboard shortcuts.

lchauvin:
@floryst I saw in an other post that you want to unify image list (DICOM + other images). I think that would be useful here, so we can select from all images which image to display on a specific view. Do you know if this is something that would be available soon ?

lchauvin:
Ok, I was trying to implement this 'multiple images' feature, but I ran into an architecture issue. If each view can have a different image, then we need to keep a mapping of which view display which images. I was working on that, but I realized this is somehow incompatible with the current architecture, where only 1 image is active at a time. So, now I'm not sure what would return `useCurrentImage` function, as we don't have a single `primaryImageID`, it depends on the view.

My idea would be to change `useCurrentImage` to take a view as argument and return the current image in this view (based on the mapping). But this is a significant change in the architecture, and to be honest, I don't know if I have the knowledge to change something that deep in the code.

Is it a path you are willing to go ? Otherwise, can you think about an alternative way ?

Thank you.

PR #578, opened 2024-04-13
fix(windowing): more robust loading of multi-component DICOMs

PaulHax:
Attached DICOM was freezing app.
[1-001.zip](https://github.com/Kitware/VolView/files/14968696/1-001.zip)

### DIFF
--- a/src/composables/useWindowingConfigInitializer.ts
+++ b/src/composables/useWindowingConfigInitializer.ts
@@ -1,12 +1,41 @@
 import { useImage } from '@/src/composables/useCurrentImage';
 import { useWindowingConfig } from '@/src/composables/useWindowingConfig';
 import { WLAutoRanges, WL_AUTO_DEFAULT, WL_HIST_BINS } from '@/src/constants';
-import { useDICOMStore } from '@/src/store/datasets-dicom';
+import { getWindowLevels, useDICOMStore } from '@/src/store/datasets-dicom';
 import useWindowingStore from '@/src/store/view-configs/windowing';
 import { Maybe } from '@/src/types';
 import type { TypedArray } from '@kitware/vtk.js/types';
 import { watchImmediate } from '@vueuse/core';
 import { MaybeRef, computed, unref, watch } from 'vue';
+
+// Original source from https://www.npmjs.com/package/compute-range and vtkDataArray
+// Modified to assume one component array
+function fastComputeRange(arr: number[] | TypedArray) {
+  const len = arr.length;
+  let min = Number.MAX_VALUE;
+  let max = -Number.MAX_VALUE;
+  let x;
+  let i;
+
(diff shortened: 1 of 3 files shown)


src/composables/useWindowingConfigInitializer.ts
```diff
@@ -30,7 +59,8 @@ function useAutoRangeValues(imageID: MaybeRef<Maybe<string>>) {
 
     // Pre-compute the auto-range values
     const scalarData = imageData.value.getPointData().getScalars();
-    const [min, max] = scalarData.getRange();
+    // Assumes all data is one component
+    const { min, max } = fastComputeRange(scalarData.getData());
```
floryst, line 63: Was the original getRange() call causing any particular issue?

PaulHax: It was working perfectly for one component, which caused NANs in downstream windowing calculations.

PaulHax: I'll try it

PaulHax: `scalarData.getRange(0)` is the default and what was happaning before with just `scalarData.getRange()`.  That is a problem when we compute the `autoRangeValues`, which computes the histogram from all components from `scalarData.getData()`.  Histogram including all compnents does not make sense to me eather.  

We are rendering all components.  Looks like a single Window and Level are getting applied to all components:
https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/OpenGL/ImageMapper/index.js#L638-L655

I see this PR as a cludge to keep the app trucking along while still sort of supporting RGB images.

floryst: In that case, this is fine. We will have to revisit this at some point.

src/composables/useWindowingConfigInitializer.ts
```diff
@@ -1,13 +1,42 @@
...
+import { getWindowLevels, useDICOMStore } from '@/src/store/datasets-dicom';
 import useWindowingStore from '@/src/store/view-configs/windowing';
 import { Maybe } from '@/src/types';
 import type { TypedArray } from '@kitware/vtk.js/types';
 import { watchImmediate } from '@vueuse/core';
 import { MaybeRef, computed, unref, watch } from 'vue';
 
+// Original source from https://www.npmjs.com/package/compute-range and vtkDataArray
+// Modified to assume one component array
+function fastComputeRange(arr: number[] | TypedArray) {
```
floryst, line 13: This function is accessible via `vtkDataArray.fastComputeRange`. You should be able to just use that + set numComponents to 1.

PaulHax: Fixed

floryst:
LGTM!

PR #588, opened 2024-04-29
Auto Segment Group by name

PaulHax:
## Automatic Segment Groups by File Name

When loading files, VolView can automatically convert images to segment groups
if they follow a naming convention. For example, an image with name like `foo.segmentation.bar`
will be converted to a segment group for a base image named like `foo.baz`.  
The `segmentation` extension is defined by the `io.segmentGroupExtension` key, which takes a
string. Files `foo.[segmentGroupExtension].bar` will be automatilly converted to segment groups for a base image named `foo.baz`. The default is `''` and will disable the feature.

This config.json will define `myFile.seg.nrrd` as a segment group for a `myFile.nii` base file.

```json
{
  "io": {
    "segmentGroupExtension": "seg"
  }
}
```

### Test dataset
Zip with config-auto-seg.json, 3DUS-Fetus.nrrd, and 3DUS-Fetus.0.segmentation.nrrd
[auto-seg-datasets.zip](https://github.com/Kitware/VolView/files/15151901/auto-seg-datasets.zip)

### DIFF
--- a/documentation/content/doc/configuration_file.md
+++ b/documentation/content/doc/configuration_file.md
@@ -49,9 +49,25 @@ VolView will include in the volview.zip file.
 }
 ```
 
-These are the supported file formats:
+Working segment group file formats:
 
 hdf5, iwi.cbor, mha, nii, nii.gz, nrrd, vtk
+
+## Automatic Segment Groups by File Name
+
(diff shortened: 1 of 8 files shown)


floryst:
LGTM

Issue #604, opened 2024-06-07
Move collapsible menu button to the toolbar and make it visible, not for mobile only


lchauvin:
I think the button for collapsing the left panel should be in the ControlStrip instead of the AppBar, and it should be visible for PCs also.

PR #605, opened 2024-06-07
feat(ControlStrip): Moved menu button to ControlStrip

lchauvin:
Related to https://github.com/Kitware/VolView/issues/604

### DIFF
--- a/src/components/App.vue
+++ b/src/components/App.vue
@@ -2,7 +2,7 @@
   <drag-and-drop enabled @drop-files="loadFiles" id="app-container">
     <template v-slot="{ dragHover }">
       <v-app>
-        <app-bar></app-bar>
+        <app-bar @click:left-menu="leftSideBar = !leftSideBar"></app-bar>
         <v-navigation-drawer
           v-model="leftSideBar"
           app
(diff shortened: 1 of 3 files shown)


floryst:
@PaulHax what do you think of the placement? The alternative would be to remove the `v-if="mobile"` and have the button showing on the left of the logo.

PaulHax:
Left of logo would look good to.  Strangely I don't have a strong opinion =)  Just a slight unease about the "connection" between the button and the drawer, but no obvious solution.  Overall I think this is a nice upgrade.

lchauvin:
I decided to put it there, because initially it was left of the logo, but I didn't see it immediately. I thought putting on the menu, you can't miss it.

floryst:
Try reverting your changes and drop the `v-if="mobile"` attr to show the menu button unconditionally and let me know what you think. I agree that there is some tenuous connection with the positioning of the menu button in the control strip, and it might work better to the left of the logo. Would be good to get your opinion on it.

lchauvin:
I could put it next to the logo if you think it's better. However, like I said, it was initially there, but it wasn't directly clear to me what was this button for, maybe because of the logo right next to it, I'm not sure. I felt like it was gonna open another dropdown menu or something like that.

floryst:
I think this is consistent with how some existing designs operate. The material design website and vuetify's website both use the triple bar icon around the logo to toggle the sidebar visibility on smaller screens. In this case, we are just enabling that functionality regardless of screen size.

Issue #610, opened 2024-06-19
Left and right arrows to change through images


lchauvin:
I think it would be easier to add keyboard shortcuts (e.g. left and right arrow) to switch between images.

PR #606, opened 2024-06-13
Incremental DICOM loading

floryst:
This PR updates VolView to support incremental DICOM loading. This is a pretty big change, so I hope to improve this issue description with details of how this is broken down.

Components:
- [x] streaming DICOM parser
- [x] incrementally-loaded image support (done, but final integration depends on a dependency)
  - [x] Depends on https://github.com/Kitware/vtk-js/pull/3077
- [x] update state file schema
- [ ] DICOMWeb streaming integration
- [x] fixing various other parts of VolView for perf and integration purposes
- [ ] cleanup remaining code TODOs

PR #611, opened 2024-06-19
feat(KeyboardShortcut): Add left and right arrow to cycle through images

lchauvin:
Related to https://github.com/Kitware/VolView/issues/610

### DIFF
--- a/src/composables/actions.ts
+++ b/src/composables/actions.ts
@@ -5,6 +5,7 @@ import { useRulerStore } from '../store/tools/rulers';
 import { usePolygonStore } from '../store/tools/polygons';
 import { Action } from '../constants';
 import { useKeyboardShortcutsStore } from '../store/keyboard-shortcuts';
+import { useDatasetStore } from '../store/datasets';
 
 const applyLabelOffset = (offset: number) => () => {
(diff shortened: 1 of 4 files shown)


PaulHax (approved):
Works good.  Thanks!

floryst:
This does have an issue where not-yet reconstructed DICOM images aren't included in the cycling. I'm unifying the image list in #606, which should make this easier.

lchauvin:
Has the list of images been unified ?

PaulHax:
@lchauvin https://github.com/Kitware/VolView/pull/606 added an imageCacheStore which holds image IDs for DICOM and other image types, even as they are loaded.  So the arrow keys probably want to loop through the `imageCacheStore.imageIds` array.

floryst:
I've given this some thought, so here are some ideas I had for this feature:
- the patient studies + images sidebar should have a backing store for it, which defines the order of the seen images. The order of the images in the image cache store might not be particularly useful, since the order still might not be what is visually seen in the sidebar
- The left/right arrow cycles will use this backing store to select the next image

lchauvin:
Should I wait for the backing store before making any modification ?

PR #615, opened 2024-06-26
feat(Camera): Synchronize image slices and cameras

lchauvin:
This feature will allow to synchronize images (i.e. slices) and cameras, so it will be easy to compare images from different modalities for a given subject for example

### DIFF
--- a/src/components/ImageDataBrowser.vue
+++ b/src/components/ImageDataBrowser.vue
@@ -15,6 +15,8 @@ import { useDatasetStore } from '../store/datasets';
 
 import { useMultiSelection } from '../composables/useMultiSelection';
 import { useLayersStore } from '../store/datasets-layers';
+import { useViewSliceStore } from '../store/view-configs/slicing'
+import { useViewCameraStore } from '../store/view-configs/camera'
 
 function imageCacheKey(dataID: string) {
   return `image-${dataID}`;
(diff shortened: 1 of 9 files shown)


lchauvin:
The only issue right now, it that it seems the camera is reset when changing images (which defeat the purpose of having camera synchronized).
I tried to comment out the `resetCamera` function is `SliceViewer` to see if it was coming from here, but it seems not.

I think it may have something to do with `disableAutoResetCamera`. I tried to turn if on (no auto reset), but then I can't see the images (not sure if they are out of the FOV or what).

PaulHax:
I'm not sure about the order VTK.js camera is updated on switching images. But maybe to diagnose: keep the existing reset stuff, (might be setting needed camera parameters not included in the "synced" ones) and add a composable/hook to trigger the position, focalpont, etc sync after the resets.

lchauvin:
I'm not sure where or how to do that. What shoud I hook for ?

PaulHax:
I'ed want to test if the problem is the timing of the camera property sync vs something else.  So could add some temp code to call the sync function on a button press, or wherever you see a camera reset.

lchauvin:
It's still unclear to me how you want to test this. What I could tell so far, is that the properties of the sync cameras are well updated. I looked in the pinia store, and I could see the values changing for all cameras when sync. However, when I change image, it resets these properties, and because images are sync, it resets to all of them.

lchauvin:
I tried to monitor the events, but all I can see when changing image is an event called `update:modelValue` being fired from `ItemGroup`. I tried to follow this event, but I don't see it calling a camera reset.

floryst:
Any status on this? Or can we close?

PR #625, opened 2024-07-16
feat(import): simpler import + state restore logic

floryst:
This commit contains two changes that probably can be split up into two separate PRs that can be applied to the base branch. I'm putting this up as a draft for now so @PaulHax can get a sense of the changes.

The pipeline code has been removed in favor of a simpler chain-of-responsibility approach, using `evaluateChain` and `asyncSelect`.

`evaluateChain` is responsible for evaluating a data source against a chain of import handlers until one of them returns a new data source.

To keep processing a data source like how the old pipeline code supported nested executions, `evaluateChain` is invoked inside a loop for every data source. `asyncSelect` is used to drive the loop execution, seleting `evaluateChain` promises whenever they are done.

The state schema is updated to generically operate on serialized data sources. Instead of special-casing for remote files, the serialized DataSource type encodes this state.


floryst:
@PaulHax FYI I'm merging these changes into streaming-base and working off of that.

Issue #538, opened 2023-12-31
Remove writeImage direction matrix workaround after upgrading @itk-wasm/image-io


PaulHax:
This code block 
https://github.com/Kitware/VolView/blob/main/src/store/segmentGroups.ts#L51-L60

will no longer be needed after this fix is incoprated in VolView:
https://github.com/InsightSoftwareConsortium/itk-wasm/pull/1017

PR #630, opened 2024-07-29
Upgrade itk-wasm

floryst:
This pulls out the ITK-wasm upgrade from the streaming branch.

* switches to using a module-scoped DICOM interface.


src/io/readWriteImage.ts
```diff
@@ -34,7 +34,8 @@ export const writeImage = async (format: string, image: vtkImageData) => {
     }
```
PaulHax, line 34: Can remove this workaround now.

Remove writeImage direction matrix workaround after upgrading @itk-wasm/image-io #538

floryst: Cool, thanks for the heads up!

PR #635, opened 2024-08-07
feat(DataSource): simplified data structure

floryst:
DataSource objects now represent a single type, rather than the bag-of-types done originally.

@PaulHax this is the PR to test (stacked on `simplified-state-load-restore` and `streaming-base`).


PaulHax:
Little bug I'm having trouble fixing: 
When I drag and drop a PET and CT series from the same study, they are automatically layered.  But I get an error:
 ```
macros2.js:50 attempt to set zero width color range
vtkErrorMacro	@	macros2.js:50
publicAPI.setMappingRange	@	ColorTransferFunction.js:981
applyColoring	@	useColoringEffect.ts:47
applyLayerColoring	@	VtkLayerSliceRepresentation.vue:97
callWithErrorHandling	@	runtime-core.esm-bundler.js:158
```
As `const imageDataRange = image.getPointData().getScalars().getRange();` returns [0, 0].  Assuming that the image is not fully loaded yet, and need to defer the color map setup for the layer until then...

floryst:
Yeah, that's something we should guard against. I suspect the PET image has slices that are completely empty, so they are initializing the range to [0, 0]. We can defer colormap application, or we can guess a "reasonable" initial data range if that's even acceptable.

floryst:
FYI I do something similar in VolumeRendering, where the opacity function editor is delayed until the entire dataset is available (because it needs to compute a histogram).

PR #653, opened 2024-09-25
Update @itk-wasm/dicom for new SEG pipeline  

### LINKED ISSUE #639: Incorrect rendering of DICOM SEG
See problematic sample below. You can download the DICOM study in question using the following:

```bash
$ pip install idc-index --upgrade
$ idc download 1.2.840.113654.2.55.192012426995727721871016249335309434385
```

https://volview.kitware.app/?urls=[s3://idc-open-data/46350591-eaf4-4f53-8790-645d72087aa2,s3://idc-open-data/3a84b4b8-b9c1-45e5-99db-d778fd5218f8,s3://idc-open-data/4c74c61b-48b1-48f4-af20-ac92b0a8a221,s3://idc-open-data/dc93f58c-2c3c-4dce-bed3-fdb291a40cee,s3://idc-open-data/9e7dc056-400c-4ac1-b01c-8e8441486f3c]

image

This is how it looks in OHIF v3:

https://viewer.imaging.datacommons.cancer.gov/v3/viewer/?StudyInstanceUIDs=1.2.840.113654.2.55.192012426995727721871016249335309434385&SeriesInstanceUIDs=1.2.840.113654.2.55.305538394446738410906709753576946604022,1.2.276.0.7230010.3.1.3.313263360.15787.1706310178.804490

image

And Slicer:

image

PaulHax:
- [x] Deal with overlapping segments coming out as multi-component images. Short putt: turn `mergeSegments` option to true, pull first output file/component for use as segment group, and pop a warning.
- [x] Stash segment label and color tags in a store on load.  Then create new segments with colors when converting SEG volume to segment group.

image

closes #639 

Suggest we just put these in an issue and not do in this PR:

- [ ] Fix thumbnail for big SEG here: https://github.com/Kitware/VolView/issues/639
- [ ] Loading indicator.  Like a spinner on the image in the image browser.
- [ ] When loading SEG volume without "background" volume, apply approate color map, surface segment group colors and labels to GUI. 
- [ ] Handle overlaping segments.


fedorov:
> Short putt: turn mergeSegments option to true and expect one component image as output.

Just to make sure there is no misunderstanding here. `mergeSegments` option will merge **non-overlapping segments** into minimal number of files. So if SEG does not have _any_ overlapping segments, unless there is a bug, you should get only one file when `mergeSegments` is on. But if some segments overlap, they will be saved into separate files with `mergeSegments` on. TotalSegmentator should not produce overlapping segments. But you can look at the NSCLC-Radiomics or QIN-Prostate-Repeatability collections that contains overlapping segments:

* [NSCLC-Radiomics](https://doi.org/10.7937/K9/TCIA.2015.PF0M9REI) example: https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.32722.99.99.203715003805996641695765332389135385095


* [QIN-Prostate-Repeatability](http://doi.org/10.7937/K9/TCIA.2018.MR1CKGND) example: https://viewer.imaging.datacommons.cancer.gov/viewer/1.3.6.1.4.1.14519.5.2.1.3671.4754.121472087445374646718121301133


@jadh4v - one of the above would be a good sample to include in testing in https://github.com/InsightSoftwareConsortium/ITK-Wasm/issues/1208.

As always, to download the images in the above example, you can use `StudyInstanceUID` available in the URL and [`idc-index` package](https://github.com/imagingdatacommons/idc-index):
```
$ pip install idc-index --upgrade
$ idc download 1.3.6.1.4.1.14519.5.2.1.3671.4754.121472087445374646718121301133
```

Note that to see overlapping segments in the above I used OHIF v2. The current OHIF v3 cannot display overlapping segments as of yet, see https://github.com/OHIF/Viewers/issues/3496.

PaulHax:
I did misunderstood.  Thanks for the clarification and files.  

Changed to do the wrong thing better.  That is: just using the first output file and popping a warning.

floryst (approved):

PR #654, opened 2024-09-26
Segment Groups: opacity slider and visibility toggle button

PaulHax:
image

Not sure this is the blend mode we want, but does give info about underlying segment groups even if full opacity.


src/composables/useSegmentGroupConfigInitializer.ts
```diff
@@ -0,0 +1,21 @@
...
+  const colorConfig = computed(() =>
+    coloringStore.getConfig(unref(viewId), unref(layerId))
+  );
+
+  watchImmediate(colorConfig, (config) => {
+    if (config) return;
+
+    const viewIdVal = unref(viewId);
+    const layerIdVal = unref(layerId);
+    coloringStore.updateConfig(viewIdVal, layerIdVal, {});
```
floryst, line 19: Is there a reason we need an initializer composable that initializes the view/layer key to an empty object?

PaulHax: The purpose of the initializer is to create the config entry for the segment group view in the coloring store.   

Passing the empty object got merged with the default config values.  Does read strange.  Added a coloringStore.initConfig() function.

floryst:
LGTM

Issue #657, opened 2024-10-03
3D view camera parameters strange after loading Fetus then Prostate


PaulHax:
Steps
1. https://volview-dev.kitware.app/
2. Load sample Fetus
3. Load sample Prostate
4. 3D camera center and clip plane? wrong:

![image](https://github.com/user-attachments/assets/a7c985e6-b808-4644-a2cb-17bebb0e3b00)

PR #658, opened 2024-10-09
fix(crop): wait for DICOM image load

floryst:
Cropping was failing on DICOM images because cropping was initialized before the DICOM data was fully loaded.


floryst:
Might fix #657

Issue #655, opened 2024-09-30
Unsupported data type error


Srujith20:
Issue Resolved

floryst:
I think there's an issue with itk-wasm opening `nii.gz` files. If you decompress them into `.nii` files, do they load?

PaulHax:
Yes, we lost the compressed `nii.gz` format support somewhere.  

After decompressing, itk-wasm returns a  `BigInt64Array` typed array, which looks like we don't handle well.  @Srujith20 Try converting the label map file to a Int32 typed array.

Srujith20:
> Yes, we lost the compressed `nii.gz` format support somewhere.
> 
> After decompressing, itk-wasm returns a `BigInt64Array` typed array, which looks like we don't handle well. @Srujith20 Try converting the label map file to a Int32 typed array.

I am encountering the following error when I try to visualize an Int32 typed array.
<img width="839" alt="Screenshot 2024-10-01 at 12 10 14 PM" src="https://github.com/user-attachments/assets/ec5b2c2d-2868-4021-8baf-ca55ac0cd688">

Srujith20:
> I think there's an issue with itk-wasm opening `nii.gz` files. If you decompress them into `.nii` files, do they load?

I am still not able to load the label after decompressing. When I try to visualize the image it opens up sometimes and sometimes it throws the "could not load" error.

PaulHax:
The loader still thinks its a Int64 image.  Check your image conversion code.  For more help here:
- Suggest using ITK's Python API.  
- Post the python code you use to do the conversion
- Post the `print(converted_image)` result
- Share the before and after image here.

PR #664, opened 2024-10-15
fix(updateFileMimeType): overwrite mimetype

### LINKED ISSUE #663: *.gz file extension failed to resolve to correct mime type
For example, when drag'n'drop a `file.nii.gz` compressed nifti file into the volview website, it cannot be opened.
image

I added some debug console log and see the `fileType` is being resolved to `application/gzip`, which is incorrect.
image
By looking up the `mimeTypes.ts` file I think it should be `application/vnd.unknown.nifti-1`.

I think the `*.gz` file names should be handled specifically in the `updateFileMimeType.ts` to correctly resolve the mime type.

Below is my quick fix for the `nii.gz` file:
image

My implementation is not very comprehensive. Hope this can be fixed by volview offical team, thanks!

floryst:
Closes #663, and maybe #655.

Overwrites/normalizes the default OS-provided mimetype. Also ensures that the mime type is a supported mime type.

@PaulHax I think this should be fine, since if it's an unsupported mime type, setting to "null" won't change any logic. 

Should also be applied here: https://github.com/Kitware/VolView/pull/635/files#diff-7d267f00b6c6e1caa0952df237d876fce197c2048228c5f5bda7dd583232a40eR10

### DIFF
--- a/src/io/import/processors/updateFileMimeType.ts
+++ b/src/io/import/processors/updateFileMimeType.ts
@@ -8,7 +8,7 @@ import { ImportHandler } from '@/src/io/import/common';
 const updateFileMimeType: ImportHandler = async (dataSource) => {
   let src = dataSource;
   const { fileSrc } = src;
-  if (fileSrc && fileSrc.fileType === '') {
+  if (fileSrc) {
     const mime = await getFileMimeType(fileSrc.file);
     if (mime) {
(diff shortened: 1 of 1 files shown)


PaulHax (approved):
I can't think of anything this would break.

PR #667, opened 2024-10-17
feat(segmentGroups): add more default segment colors

PaulHax:


floryst (approved):

PaulHax:
> Looks like there is some type changes that causes the build to fail.


Fixed by adding visible prop.

floryst (approved):

PR #669, opened 2024-10-17
Segments:  opacity slider and visibile toggle

PaulHax:
image

### DIFF
--- a/src/store/segmentGroups.ts
+++ b/src/store/segmentGroups.ts
@@ -26,7 +26,6 @@ import {
 import { FileEntry } from '../io/types';
 import { ensureSameSpace } from '../io/resample/resample';
 import { useDICOMStore } from './datasets-dicom';
-import { vi } from 'vitest';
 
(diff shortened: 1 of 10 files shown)


src/store/segmentGroups.ts
```diff
@@ -26,6 +26,7 @@ import {
 import { FileEntry } from '../io/types';
 import { ensureSameSpace } from '../io/resample/resample';
 import { useDICOMStore } from './datasets-dicom';
+import { vi } from 'vitest';
```
floryst: Unused import

floryst (approved):
LGTM with one unused import comment.

I re-ran the test in case it's transient, but if it fails again we'll take a look.

src/store/segmentGroups.ts
PaulHax: got it!

PaulHax:
Yes, the e2e layer test got flakey on ubuntu couple weeks ago.

PR #670, opened 2024-10-21
Outline segment group rendering

PaulHax:
image

### DIFF
--- a/src/components/vtk/VtkSegmentationSliceRepresentation.vue
+++ b/src/components/vtk/VtkSegmentationSliceRepresentation.vue
@@ -135,6 +135,28 @@ const applySegmentColoring = () => {
 
 watchEffect(applySegmentColoring);
 
+sliceRep.property.setUseLabelOutline(true);
+sliceRep.property.setUseLookupTableScalarRange(true); // For the labelmap is rendered correctly
+
+// watchEffect(() => {
+//   sliceRep.property.setLabelOutlineOpacity(opacity.value);
+// });
+
+const outlinePixelThickness = 2;
+watchEffect(() => {
+  if (!metadata.value) return; // segment group just deleted
+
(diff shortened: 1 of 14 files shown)


PaulHax:
Looks like this vtk.js commit broke label map rendering when the labelmap image is `Int16Array`

https://github.com/Kitware/vtk-js/commit/b04e59d1a10e5a6a06632799494a395ddcbfb2af#diff-18c5d83bebc14325d84e520000b95e06dadffa088fe27eef138af64744171b89R1891

PaulHax:
@floryst This is good for review now.

src/components/vtk/VtkSegmentationSliceRepresentation.vue
```diff
@@ -135,6 +138,35 @@ const applySegmentColoring = () => {
 
 watchEffect(applySegmentColoring);
 
+const configStore = useSegmentGroupConfigStore();
+const config = computed(() =>
+  configStore.getConfig(unref(viewId), unref(segmentationId))
+);
+
+const outlineThickness = computed(() => config.value?.outlineThickness ?? 2);
+// @ts-expect-error vtk.js types are incomplete
```
floryst, line 147: Can you please make a PR against vtk.js to fix these types?

floryst (approved):

PaulHax:
> I'm wondering if we should move some of the opacity sliders to a menu. For instance, the segment opacity slider could be moved into the edit box.

Maybe a yes, and for that. Slicer segmentations opacity controls:

image

src/components/vtk/VtkSegmentationSliceRepresentation.vue
PaulHax: https://github.com/Kitware/vtk-js/pull/3174

PR #673, opened 2024-10-23
feat(segmentGroups): support overlapping segments in SEG files

### LINKED ISSUE #672: Support overlaping segments from SEG dicom file
Approach: make a new segment group for each component output from  `@itk-wasm/dicom::readOverlappingSegments`.  Could be very simple by just looping code here: https://github.com/Kitware/VolView/blob/main/src/store/segmentGroups.ts#L253-L294

PaulHax:
Each component in the image to convert is a non overlapping labelmap. Create another segment group for each component of the image to convert.

closes #672


floryst (approved):

floryst:
I'm assuming the failing tests are transient

PR #679, opened 2024-10-29
feat(Slice): Use keyboard to change slices of the active view

lchauvin:
This needs this PR to be merged first: https://github.com/Kitware/VolView/pull/676

### DIFF
--- a/src/composables/actions.ts
+++ b/src/composables/actions.ts
@@ -39,20 +39,12 @@ const showKeyboardShortcuts = () => {
   keyboardStore.settingsOpen = !keyboardStore.settingsOpen;
 };
 
-const nextSlice = () => () => {
+const changeSlice = (offset: number) => () => {
   const { currentImageID } = useCurrentImage();
   const { activeViewID } = useViewStore();
 
   const { slice: currentSlice } = useSliceConfig(activeViewID, currentImageID);
-  currentSlice.value += 1;
-};
-
-const previousSlice = () => () => {
-  const { currentImageID } = useCurrentImage();
-  const { activeViewID } = useViewStore();
-
-  const { slice: currentSlice } = useSliceConfig(activeViewID, currentImageID);
-  currentSlice.value -= 1;
+  currentSlice.value += offset;
 };
 
 export const ACTION_TO_FUNC = {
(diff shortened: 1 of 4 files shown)


src/composables/actions.ts
```diff
@@ -36,6 +39,22 @@ const showKeyboardShortcuts = () => {
...
+  currentSlice.value += 1;
+};
+
+const previousSlice = () => () => {
+  const { currentImageID } = useCurrentImage();
+  const { activeViewID } = useViewStore();
+
+  const { slice: currentSlice } = useSliceConfig(activeViewID, currentImageID);
+  currentSlice.value -= 1;
+};
```
PaulHax, line 56: To remove duplicate code, could define something like `offsetSlice(offset: number)` then
```
const nextSlice = () => offsetSlice(1)
const previousSlice = () => offsetSlice(-1)
```

src/components/LayoutGrid.vue
```diff
@@ -28,6 +29,13 @@ import { useViewStore } from '../store/views';
 
 export default defineComponent({
   name: 'LayoutGrid',
+  methods: {
+    onFocusView(id: string, type: string) {
```
PaulHax, line 33: 👍️ Good

Also, when I scroll with mouse wheel on a view, then press the arrow keys, its not always the same view that gets its slice changed.  So maybe setActiveView on scroll events too?

PaulHax (commented):
Making dreams come true here.

If you rebase on top of `main` rather than merge we can avoid thoes pesky merge commits.

src/components/LayoutGrid.vue
lchauvin: I'm trying to fix this but I'd like to check with you which behavior you would prefer.

For now, scrolling with the mouse will scroll the image under the cursor when scrolling instead of the active view.

I could either scroll the active view, wherever the cursor is, or I could change the active view to the view under the cursor when scrolling with the mouse.

What do you think would be best ?

lchauvin: I cannot really find where this scroll event is handled in the code. Could you point me in the right direction please ?

PaulHax: Yes, somewhere in VTK.js code I think.  So we just need to add our own mouse listener (that possibly has `capture=true`) which does not consume event.

PaulHax: There may be some more general method to register a callback on the vtk.js manipulator, and whever the manipulator does something on the view (pan, zoom, slice scroll) we set the active view.  But I think we cover the pan and zoom cases already with our mouse button callback...  Your call =)

lchauvin: I fix the scrolling to set the active view id also. Let me know if it works for you. I also refactored some code, like you suggested.

PaulHax (approved):
Look great feels, feels great.  Been wanting this!

floryst (approved):
LGTM! Thanks for adding this!

PR #695, opened 2025-01-28
chore: upgrade webdriverio to v9

floryst:
Upgrades the visual testing framework, which also removes the hard dependency on the canvas dep.


PaulHax (approved):
very nice.

PR #705, opened 2025-02-21
Batch edit segment groups

PaulHax:
image


floryst:
Right now, when I delete all segment groups a new one is automatically created. I think we should revert to the empty state instead of auto-creating a new segment group.

PaulHax:
> Right now, when I delete all segment groups a new one is automatically created. I think we should revert to the empty state instead of auto-creating a new segment group.

Fixed

floryst:
It's working well! One last suggestion: when there are no segment groups, I think hiding the batch selection controls would look nice.

PaulHax:
> when there are no segment groups, I think hiding the batch selection controls would look nice.

Done

floryst (approved):

PR #706, opened 2025-02-21
feat: Keyboard shortcut to delete current image and clear scene

lchauvin:
Add a keyboard shortcut to delete the current image (ctrl+d), and clear the whole scene (ctrl+w)

### DIFF
--- a/src/components/ControlButton.vue
+++ b/src/components/ControlButton.vue
@@ -24,7 +24,7 @@
 
 <script>
 export default {
-  name: 'ToolButton',
+  name: 'ControlButton',
   props: {
     icon: { type: String, required: true },
     name: { type: String, required: true },
--- a/src/components/ControlsStripTools.vue
+++ b/src/components/ControlsStripTools.vue
@@ -11,7 +11,7 @@
     >
       <menu-control-button
         icon="mdi-circle-half-full"
-        name="Window & Level"
+        :name="`Window & Level [${nameToShortcut['Window & Level']}]`"
         :active="active"
         :disabled="noCurrentImage"
         @click="toggle"
@@ -22,7 +22,7 @@
     <groupable-item v-slot:default="{ active, toggle }" :value="Tools.Pan">
       <control-button
         icon="mdi-cursor-move"
-        name="Pan"
+        :name="`Pan [${nameToShortcut['Pan']}]`"
         :buttonClass="['tool-btn', active ? 'tool-btn-selected' : '']"
         :disabled="noCurrentImage"
         @click="toggle"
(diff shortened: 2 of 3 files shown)


src/config.ts
```diff
@@ -287,6 +287,9 @@ export const ACTION_TO_KEY = {
   decrementLabel: 'q',
   incrementLabel: 'w',
 
+  deleteCurrentImage: 'ctrl+d',
+  clearScene: 'ctrl+w',
```
PaulHax, line 291: These combos are used by Chrome too.  Any other ideas for defaults here?

https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Ctab-window-shortcuts

PaulHax (approved):
Looks good!  Gracias.

src/config.ts
lchauvin: I was thinking that without a modifier key maybe people could press on it by accident (to write something) and close the scene or delete an image by accident

PaulHax: hmm, your right, accidential would be sad.

lchauvin: I put 'd' and 'k' for now. Do you want me to change for  'ctrl+/' and 'ctrl+.' ?

PaulHax: No no, I thrashed around enought.  Lets see what @floryst thinks.

floryst: I would rather have deletion shortcuts behind a modifier. This can either be ctrl or even just shift (e.g. `SHIFT-3`).

floryst (changes requested):
Whatever your final keybinds are, please add them to https://github.com/Kitware/VolView/blob/main/docs/mouse_controls.md.

src/config.ts
lchauvin: I think I would go for ctrl instead. I'm thinking that Shift-1, etc... could be use to setup some presets later on (like Axial primary, Coronal primary, etc...)

lchauvin:
> Whatever your final keybinds are, please add them to https://github.com/Kitware/VolView/blob/main/docs/mouse_controls.md.

I changed the shortcut to add modifier, and modified mouse_controls.md

src/composables/actions.ts
```diff
@@ -47,6 +48,19 @@ const changeSlice = (offset: number) => () => {
...
 };
 
+const clearScene = () => () => {
+  const datasetStore = useDatasetStore();
+  datasetStore.removeAll();
+};
+
+const deleteCurrentImage = () => () => {
+  const datasetStore = useDatasetStore();
+  datasetStore.remove(datasetStore.primaryImageID!);
```
floryst, line 58: I'm all for updating `datasetStore.remove` so that we can pass in null image IDs without checking. That way we can drop the `!` operator here.

```
remove(id: string | null) {
  if (!id) return
  ...
}
```

lchauvin: I added the changes

PaulHax:
I was trying to squash some of lchauvin's commits and rebase rather than merge on main... but messed up force pushing them back on this PR =/  

Guess we just use #713

PR #713, opened 2025-03-06
Delete image shortcut

PaulHax:
supersedes #706


PaulHax:
@lchauvin Good to go?

PR #715, opened 2025-03-11
feat: add scalar probe tool

PaulHax:


src/components/ProbeView.vue
```diff
@@ -0,0 +1,55 @@
...
+import { storeToRefs } from 'pinia';
+
+const probeStore = useProbeStore();
+const { probeData } = storeToRefs(probeStore);
+
+const formattedProbeItems = computed(() => {
+  if (!probeData.value) return [];
+  const sampleItems = probeData.value.samples.map((sample) => ({
+    label: sample.name,
+    value: sample.displayValue.join(', '),
```
floryst, line 13: We should shorten the displayValue, since floats will show up as very long numbers.

Something like the following:
```
if Number.isInteger(value) return value
if 0 < Math.abs(value) < 1 return value.toExponential(2)
return value.toFixed(2)
```

PaulHax: Yes.  I did this with `num.toPrecision` which is similar.  `toPrecision` won't work well for images with large `origin` values though.  

image

floryst: I think toPrecision isn't the right thing to use here, because of the issue you mentioned and because we're not looking to show fixed-precision values. toFixed is a better approach, but truncates to 0 for sufficiently tiny values in [0,1].

PaulHax: Ok, added a `shortenNumber` func with your code.

floryst: What about for the pixel value? 
image

floryst: To clarify, I was thinking of ways to shorten the pixel value, not the position. I think using `Math.round` for the position is good, since we don't need much more resolution beyond the decimal point.

PaulHax: Got it now.

1isten:
@PaulHax the `ScalarProbe.vue` causes console warning:
image

can we add `<template><slot></slot></template>` to the file to eliminate that vue warning?
image

PR #722, opened 2025-03-20
Fix vue warning and potential js error in ScalarProbe.vue

1isten:
the `ScalarProbe.vue` causes console warning:
image

and causes this js error when switching Layout, for example from Quad View to Axial Only:
image
image

this PR fix above issues

### DIFF
--- a/src/components/tools/ScalarProbe.vue
+++ b/src/components/tools/ScalarProbe.vue
@@ -148,3 +148,5 @@ watch([currentImageID, sampleSet], () => {
   probeStore.clearProbeData();
 });
 </script>
+
+<template><slot></slot></template>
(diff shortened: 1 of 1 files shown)


PaulHax:
This is awesome, thank you so much for fixing my bugs.  To get this mergred would you mind

- Moving the undefined check for the missing `sample.rep` to the `getBaseSlice()` func?  The base slice "sample" is basicly invalid if the `ref` is not defined yet, so better to catch that earler in the reactive chain.
- Change the commit messages to follow semantic commit style.  So change them to `fix(ScalarProbe): eliminate vue console warning` and `fix(ScalarProbe): avoid accessing undefined rep`
- Be sure to format the code style so Pretter is happy: `npm run prettify`

1isten:
Thanks @PaulHax , I created a new PR: https://github.com/Kitware/VolView/pull/723, can you check?

PR #723, opened 2025-03-20
Fix vue warning and avoid accessing undefined rep in ScalarProbe.vue

1isten:
this is an adjustment version for PR https://github.com/Kitware/VolView/pull/722


PaulHax (approved):
Looks great.  Thank you.

PR #732, opened 2025-04-28
fix(seg): regression with loading seg images

floryst:
Loading segmentations how happens through the streaming interface.


PaulHax (approved):
Turned out to be simpler than I though it would be!

PR #736, opened 2025-05-06
fix(cachedStreamFetcher): avoid double download

PaulHax:
When the entire stream finished but had no Content-Length header, treat full cache as total length.

Girder does not set Content-Length header in its response and we were fetching the entire stream twice.


floryst (approved):
Thanks for catching that!

PR #738, opened 2025-05-12
fix(windowing): no auto window level if user set already

PaulHax:
Keeps the tag or auto histogram window level values from taking over if user already used range manipulator

Adds image-stats store to hold image range and histogram computations

- [x] switching images loses window level if auto histogram value picked
- [x] Radio group tag/ct-presets/auto-range GUI does not update corretly if range manipulator used
- [x] ~~Disable out auto histogram radio options if not fully loaed yet?~~  Or update when histogram is computed?
- [x] ~~fix MouseRangeManipulator step/sensativity computation for windowing~~  actually messing with step is not the correct parameter and finding good scale parameter is not obvious.
- [x] fix deserialization of windowing config from state file

### DIFF
--- a/src/composables/useWindowingConfig.ts
+++ b/src/composables/useWindowingConfig.ts
@@ -1,4 +1,4 @@
-import useWindowingStore from '@/src/store/view-configs/windowing';
+import { useWindowingStore } from '@/src/store/view-configs/windowing';
 import { Maybe } from '@/src/types';
 import type { Vector2 } from '@kitware/vtk.js/types';
 import { MaybeRef, unref, computed } from 'vue';
--- a/src/composables/useWindowingConfig.ts
+++ b/src/composables/useWindowingConfig.ts
@@ -15,13 +15,13 @@ export function useWindowingConfig(
     if (!imageIdVal) return undefined;
     const viewIdVal = unref(viewID);
     if (!viewIdVal) return undefined;
-    return store.getConfig(viewIdVal, imageIdVal);
+    return store.getConfig(viewIdVal, imageIdVal).value;
   });
 
   const generateComputed = (prop: 'width' | 'level') => {
     return computed({
       get: () => {
-        return config.value?.value?.[prop] ?? 0;
+        return config.value?.[prop] ?? 0;
       },
       set: (val) => {
         const imageIdVal = unref(imageID);
--- a/src/composables/useWindowingConfig.ts
+++ b/src/composables/useWindowingConfig.ts
@@ -2,12 +2,14 @@ import useWindowingStore from '@/src/store/view-configs/windowing';
 import { Maybe } from '@/src/types';
 import type { Vector2 } from '@kitware/vtk.js/types';
 import { MaybeRef, unref, computed } from 'vue';
+import { useImageStatsStore } from '@/src/store/image-stats';
 
 export function useWindowingConfig(
   viewID: MaybeRef<string>,
   imageID: MaybeRef<Maybe<string>>
 ) {
   const store = useWindowingStore();
+  const imageStatsStore = useImageStatsStore();
   const config = computed(() => store.getConfig(unref(viewID), unref(imageID)));
 
   const generateComputed = (prop: 'width' | 'level') => {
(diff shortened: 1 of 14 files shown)


src/composables/useWindowingConfig.ts
```diff
@@ -1,30 +1,43 @@
...
     return computed({
-      get: () => config.value?.[prop] ?? 0,
+      get: () => {
+        return config.value?.[prop] ?? 0;
+      },
       set: (val) => {
         const imageIdVal = unref(imageID);
         if (!imageIdVal || val == null) return;
-        store.updateConfig(unref(viewID), imageIdVal, { [prop]: val });
+        store.updateConfig(unref(viewID), imageIdVal, { [prop]: val }, true);
```
floryst, line 29: Not sure what this `true` parameter is doing here. Might this be accidentally added?

src/composables/useWindowingConfigInitializer.ts
```diff
@@ -96,94 +27,52 @@ export function useWindowingConfigInitializer(
...
-        },
+        width: firstTagVal.width,
+        level: firstTagVal.level,
       });
     }
 
-    const forcedWL = store.runtimeConfigWindowLevel;
-    if (forcedWL) {
+    const jsonWidthLevel = store.runtimeConfigWindowLevel;
+    if (jsonWidthLevel) {
```
floryst, line 50: nit: just name this as `widthLevel`. Prefixing with `json` makes it seem like the type is a string containing json, when it's just an object.

src/store/image-stats.ts
```diff
@@ -0,0 +1,223 @@
...
+    };
+  };
+
+  const internalRemoveStats = (imageID: string) => {
+    delete stats[imageID];
+  };
+
+  const setupImageWatcher = (id: string) => {
+    if (scalarRangeWatchers[id]) {
+      scalarRangeWatchers[id]();
```
floryst, line 108: Question: if the image watcher already exists, do we want to set it up again or keep the existing one?

If the latter, then a nit: add a comment here specifying that this is a stop function

src/store/image-stats.ts
```diff
@@ -0,0 +1,223 @@
...
+      ...(stats[imageID] ?? { scalarMin: 0, scalarMax: 0 }),
+      autoRangeValues: autoValues,
+    };
+  };
+
+  const internalRemoveStats = (imageID: string) => {
+    delete stats[imageID];
+  };
+
+  const setupImageWatcher = (id: string) => {
```
floryst, line 106: This function has multiple watchers. I would suggest using an effectScope so you can clean everything up at once.

src/composables/useWindowingConfigInitializer.ts
PaulHax: Fixed.

src/composables/useWindowingConfig.ts
PaulHax: Thats the "user did this", don't reset when fully loaded flag.

src/store/image-stats.ts
PaulHax: I think once is enough =)

src/store/image-stats.ts
PaulHax: Added an EffectScope in the code that calls this function.

src/composables/useWindowingConfig.ts
floryst: Ah, I missed that somehow. Thanks!

src/store/image-stats.ts
```diff
@@ -0,0 +1,222 @@
...
+          }
+        })
+        .catch((error) => {
+          console.error(
+            `[ImageStatsStore] Auto range computation for image ${id} FAILED:`,
+            error
+          );
+          messageStore.addError(
+            `Auto range computation failed for image ${id}`,
+            error instanceof Error ? error : String(error)
```
floryst, line 146: minor improvement: You can use `ensureError(error)` here

PaulHax: fixed

PR #742, opened 2025-05-21
fix(VtkViews): reset clipping planes on view layout change

PaulHax:


PaulHax:
> Is disableCameraAutoReset used anywhere else?

onlyin VtkSliceView and VtkVolumeView

> Does the 3D view exhibit this bug?

It did, already fixed

floryst (commented):
Overall looks good. Just two comment-related requests:
- You can drop the comments added in this PR, since they read like note-to-self messages
- Can you please add a comment to `disableCameraAutoReset` in the viewCameraStore that says it's used for disabling calls to resetCamera but not resetCameraClippingRange?

PaulHax:
I addressed the code comment requests.

floryst (approved):

PR #743, opened 2025-05-21
feat(crosshairs): add hold key for temp crosshairs tool

PaulHax:


src/components/ControlsStripTools.vue
```diff
@@ -180,6 +180,15 @@ export default defineComponent({
...
     });
 
+    const keys = useMagicKeys();
+    const enableTempCrosshairs = computed(
+      () => keys[actionToKey.value.temporaryCrosshairs].value
+    );
+    watch(enableTempCrosshairs, (enable) => {
+      if (enable) toolStore.activateTemporaryCrosshairs();
+      else toolStore.deactivateTemporaryCrosshairs();
+    });
```
PaulHax, line 190: should this setup go here, or perhaps in actions.ts?

src/config.ts
```diff
@@ -270,6 +270,7 @@ export const ACTION_TO_KEY = {
   paint: 'p',
   rectangle: 'r',
   crosshairs: 'c',
+  temporaryCrosshairs: 'v',
```
PaulHax, line 273: Better default key idea?

floryst:
I forgot to submit my review.

I'm wondering if we should special-case disable the temp crosshairs when in crosshairs mode. Nothing breaks, so I'm fine leaving it as-is.

src/config.ts
PaulHax: Changed to shift-c

PR #745, opened 2025-05-22
fix(dicom): load images with NumberOfFrames == 1

### LINKED ISSUE #744: AllocateImageFromChunks Found a multi-frame chunk in a group of chunks of size > 1 bug
Hitting this line with a DICOM

https://github.com/Kitware/VolView/blob/main/src/utils/allocateImageFromChunks.ts#L93-L98

floryst:
Closes #744 

DICOM images can have NumberOfFrames set to "1".

Throw an error when encountering a series of DICOM slices where the first one is multi-frame, since we don't handle those right now (and I have no test data to validate against).


PaulHax (approved):
Better.  Maybe we leave the issue open or make another to deal with these DICOM series types?

floryst:
I'm sure such a series of DICOM exists, but as we don't have test data for it I think we can close it for now with a comment saying that we will re-open if this comes up again in the future.

PR #747, opened 2025-05-27
feat(segments): rasterize polygons as segments

floryst:
Changes:
- Disabling the paint tool will not deselect the active segment group
- Polygons can now be rasterized to any given segment in a given segment group

image


PaulHax (approved):
Works great.

PR #755, opened 2025-06-09
feat(paint): thresholding support

floryst:
Adds thresholding capabilities to painting.

image

### DIFF
--- a/src/components/PaintControls.vue
+++ b/src/components/PaintControls.vue
@@ -92,7 +92,7 @@
               @input="setMinThreshold($event.target.value)"
               variant="underlined"
               class="mt-n3 pt-0 pl-2 opacity-70"
-              style="max-width: 80px"
+              style="width: 80px"
               density="compact"
               hide-details
               type="number"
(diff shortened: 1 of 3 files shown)


src/components/PaintControls.vue
```diff
@@ -51,73 +51,128 @@
...
+      <v-row no-gutters class="mb-1">Threshold </v-row>
+      <v-row v-if="currentImageStats" no-gutters align="center">
+        <v-range-slider
+          v-model="threshold"
+          :min="currentImageStats.scalarMin"
+          :max="currentImageStats.scalarMax"
+          :step="thresholdStepGranularity"
+        >
+          <template #prepend>
+            <v-text-field
```
PaulHax, line 90: With the prostate, as I drag down the top handle, the text fields widths' jerks wider/narrower a tiny bit.  Happens when the appened text value drops a magnatude.

PaulHax (approved):
Pretty sweet.

Switching between images can be a bit strange though as last images min/max is kept...

floryst:
I pushed changes to fix the styling and updating the threshold on image switch.

I can see reasoning for preserving the threshold values for each image. @PaulHax what do you think?

PaulHax:
> I can see reasoning for preserving the threshold values for each image. @PaulHax what do you think?

That makes sense, but i'm good with punting and tackling in "multi-image multi-view" work.

PR #757, opened 2025-06-24
Gaussian smoothing process for a segment 

PaulHax:
image


floryst (commented):
I'm still not 100% set on the workflow for choosing between the original and processed (something like "apply/cancel" with preview toggles makes more sense to me), but I don't have a better idea at the moment.

src/components/ProcessWorkflow.vue
```diff
@@ -0,0 +1,94 @@
...
+const showingOriginal = computed(() => processStore.showingOriginal);
+
+function startCompute() {
+  const id = paintStore.activeSegmentGroupID;
+  if (!id) return;
+  processStore.startProcess(id, props.algorithm);
+}
+
+function handleToggleChange(value: number | string) {
+  const numValue = typeof value === 'string' ? parseInt(value, 10) : value;
```
floryst, line 80: Would this just not be a number always? If not, could just do `Number(value)`

floryst (commented):
Overall I think this is just about ready to go.

src/components/ProcessWorkflow.vue
PaulHax: fixed to 
```
function handleToggleChange(value: number) {
  const shouldShowOriginal = value === 0;
```

floryst (approved):
Nice Job! I'm good with these changes.

PR #760, opened 2025-07-03
feat(segments): add segment locking with paint protection

TomTonroe:
Adds lock/unlock buttons to individual segments to prevent them from accidental modification when painting other segments.

image


floryst (approved):
Thanks for the feature addition!

PR #764, opened 2025-07-11
Fix Python Server for new Store shape

PaulHax:


src/store/image-cache.ts
```diff
@@ -107,6 +107,21 @@ export const useImageCacheStore = defineStore('image-cache', () => {
...
 
+  /**
+   * Updates an existing image's VTK data while maintaining the same ID.
+   */
+  function updateVTKImageData(id: string, newImageData: vtkImageData): void {
+    const progressiveImage = imageById[id];
+    const oldImageData = progressiveImage.vtkImageData.value;
+
+    progressiveImage.vtkImageData.value = newImageData;
+    newImageData.modified();
```
floryst, line 118: Best to add a comment here as to why we need a `modified()` call. Also, should we be calling `modified()` on the off-chance that `oldImageData === newImageData`?

PR #767, opened 2025-07-31
feat(paint): add cross-plane sync during painting

TomTonroe:
- Adds cross-plane synchronisation and camera centering of non-active views during painting.
- Toggle added to the settings panel (off by default).


TomTonroe:
> I like this feature. Kind of want it give it top billing by putting the switch up in the upper left side panel paint controls.

Thanks! I have moved the toggle to the top left of the paintControls panel.

PR #770, opened 2025-08-08
chore: upgrade WebDriverIO to v9 and fix TypeScript errors

PaulHax:
The supporting webdriverio pacakges were not on the same version as the main package.


floryst:
LGTM!

PR #774, opened 2025-08-12
Multi-series viewing

floryst:
## Overview
Adds support for multi-series viewing. This fundamentally changes how VolView is used, so a lot of changes were necessary.

Underpinning this change are two core interfaces: how layouts/views are accessed and updated, and the useCurrentImage hook.
- views are no longer fixed, since the number of views can now be changed as well as updated. Any code that needs to reference the view properties must now use the API available through the view store
- useCurrentImage depends on the active/selected view's loaded image

Additionally, this whole change is enabled from vtkRenderWindow supporting parenting to avoid exhausting WebGL contexts.

@PaulHax I'm tapping you to review functionality in the preview deployment. Deprioritize code review (unless you have bandwidth), since it is unfortunately large.

## Remaining tasks
- [x] address painting functionality
- [x] show viewed dataset name in the view itself
- [x] Fix Sync painting mode
- [x] Fix typescript build erros
- [x] serializaton saving and loading of session.volview.zips
- [x] Fix Rendering tab to layers and the rendering controls if 3d in view.
- [x] Fix e2e tests.
- [x] Fix layers.
- [x] Backwards compatablity of volview.zip
- [x] Make sure view config works with Yaml.
- [x] Test view maximing with double click
- [x] Fix oblique views
- [x] Maintain paint brush size per image.  Paint on one image, then paint on another image, paint on first image, verify paint brush size is the same as the first stroke.
- [x] Check for commented out code
- [x] Simplify view changing dropdown
- [ ] Make a video
- [x] update docs
- [ ] ~Document in gh issue of segment groups and mearements being invislibe in UI if there base image view is not selected.~
- [x] Menu option to apply a volume to all viewports with a single click. 

## Non-exhaustive list of changes:

commits
- Use a vtkRenderWindow parent for single-context rendering
- Simplify resize-to-fit auto-fit state tracking using pointerdown/up

refactors
- useSyncWindowing is now used to synchronize windowing configs across views
- LayoutDirection dropped in favor of using "H" and "V"
- Layout type changed
- useCurrentImage() now relies on activeView -> dataID
- primaryImageID and primaryDataset is removed from dataset store
- use of events for synchronization of state
- layout configuration now supports updating the view type associated
with a layout slot

fixes
- datasets calls removeData on more stores

file changes
- windowing.ts
  - getConfig no longer returns computed
  - syncAcrossViews is removed in favor of useSyncWindowing
- image-stats.ts
  - getAutoRangeValues no longer returns computed
  - add removeData
- WindowLevelControls.vue
  - Use activeView instead of the first view ID
- doubleKeyRecord.ts
  - Add forEachEntry
- useWindowingConfigInitializer.ts
  - drop the config initializer in favor of computing the defaults in the windowing store
- SliceView.vue
  - fix camera initialization when reloading a dataset
- useSlicingConfigInitializer.ts
  - drop the initializer in favor of computing the defaults in the slicing store


src/components/LayoutGrid.vue
```diff
@@ -1,46 +1,40 @@
...
-        <component
-          :is="item.component"
-          :key="item.id"
-          :id="item.id"
-          :type="item.viewType"
-          v-bind="item.props"
-          @focus="onFocusView(item.id!, item.viewType!)"
-        />
-      </div>
+      <LayoutGridItem
```
PaulHax, line 5: Entry point for views

src/components/ObliqueViewer.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <div class="d-flex flex-row flex-grow-1">
+    <div class="d-flex flex-column flex-grow-1">
+      <div class="d-flex flex-grow-1" style="border-bottom: 1px solid gray">
```
PaulHax, line 4: could be css grid or flex equal to fix 1 pixel offset

src/store/views.ts
```diff
@@ -1,149 +1,367 @@
...
-        this.restoreLayout();
-      } else {
-        this.maximizeView(viewID);
+  const layout = ref<Layout>(DefaultLayout);
+  // which assigns view IDs to layout slots
+  const layoutSlots = ref<string[]>([]);
+  const viewByID = reactive<Record<string, ViewInfo>>({});
+  const activeView = ref<Maybe<string>>();
+
+  const isActiveViewMaximized = ref(false);
```
PaulHax, line 78: Layout view state

PR #778, opened 2025-08-19
fix(config): Add sample data visibility env flag

### LINKED ISSUE #759: Fix seeing "Sample Data Browser" ever if configured to visiblity false
We can turn off visibilty of the Sample Data section in the data broser via a runtime config JSON.  But, the sample section is still shown for a bit when while the browser is blocked on initial URL manifest processing.

arhowe00:
Closes #759

Configures the initial visibility state at build time. Can prevent a benign race condition between the default (true) and runtime-fetched (false) settings where delayed state update looks ugly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,3 +14,6 @@ VITE_ENABLE_REMOTE_SAVE=true
 
 # VolView server remote URL
 VITE_REMOTE_SERVER_URL=http://localhost:4014
+
+# Controls the initial visibility of the Sample Data section.
+VITE_SHOW_SAMPLE_DATA=true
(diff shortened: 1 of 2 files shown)


PaulHax:
lets remove the non working `configJson.ts` code that trys to configure this

arhowe00:
> lets remove the non working `configJson.ts` code that trys to configure this

I thought we still wanted the ability to override the `.env` value with the json configuration file. Do you want the configuration file to still work to enable/disable it?

floryst:
You'll need to add `VITE_SHOW_SAMPLE_DATA` to https://github.com/Kitware/VolView/blob/main/.github/workflows/checks.yml in order for the tests to pass.

Issue #773, opened 2025-08-12
3D Rendering is too blur when rotating


phtran-dev:
Hi All,
I am doing trial testing volview on https://volview.kitware.app. My desktop has integrated GPU onboard. My CT series has 512 images, when I  rotate the Volume, it's very blur. Do you have any suggestion or setting or minimize the blurness ?

Thanks

phtran-dev:
Thanks @floryst , I tried 3D Slicer on the same machine but it works very well. There is no blurriness when rotating the volume. There is some discussion about the blurriness on discourse, i.e , https://discourse.vtk.org/t/in-volume-rendering-how-to-set-the-rendering-result-to-not-blur-when-rotating-the-camera/11294/2 . And 3D slicer exactly uses the vtkGPUVolumeRayCastMapper to mitigate the issue. Can Volview apply the same Mapper to prevent the blurriness ? 

<img width="1921" height="1029" alt="Image" src="https://github.com/user-attachments/assets/1f2ad5f7-bdb4-4dcd-af2f-8791377afee0" />

PaulHax:
We stopped manually doing downsampling in [#779](https://github.com/Kitware/VolView/pull/779) and now rely on VTK.js automatic downsampling which is less aggressive.

PR #779, opened 2025-08-20
fix(rendering): Removes downsampling of the volume

sankhesh:
Related to #773


floryst:
Ok, I'm not sure why the tests are failing at the moment. I've re-run them several times...

PR #794, opened 2025-10-01
fix(server): Await web socket room entry (must be async)

arhowe00:
Some of the io-bound functions in the rpc server are async and must be awaited.

## Before:
1. Enable the server by adding the path to `VITE_REMOTE_SERVER_URL` to `.env`
2. Go to "REMOTE" module
3. Choose numbers (e.g. "3" and "5") to add and click "=>"
4. The backend never returns the result with the following message:

```
/Users/andrew.howe/projects/vvi/2025-nvidiagtcdc-vvi/server % poetry run python -m volview_server -P 4014 ./examples/example_api.py
The "poetry.dev-dependencies" section is deprecated and will be removed in a future version. Use "poetry.group.dev.dependencies" instead.
======== Running on http://localhost:4014 ========
(Press CTRL+C to quit)
coroutine 'AsyncServer.enter_room' was never awaited
```

## After:

1. Apply patch
2. Repeat steps 1-3 from above
3. The addition above returns "8" within the app, with no log messages in backend

## Note:

The deps for the python server might need to be upgraded but that's more trivial so I didn't push the changes I made to pyproject.toml as I wasn't sure if they were the right choices.


arhowe00:
> If you want to include the Python server deps changes you know works, that would be great.

I was able to resolve this with a minimal numpy upgrade to latest, let me know if things work with that.

PR #796, opened 2025-10-07
fix(config): Sample data toggle

zachmullen:
This supersedes #778 (just fixing the remaining issues on that PR). I opened a new PR for it since the old one was on a private fork.


PaulHax (approved):
Thank you.

PR #797, opened 2025-10-07
feat(test): upload screenshots of full browser window on any E2E test failure

zachmullen:
Don't merge this yet, I'm just trying to test to make sure this works in GH Actions.


zachmullen:
Based on the failure here, it looks like what's happening when this test flakes out is that the browser is crashing. My hypothesis would be it's getting SIGKILLed due to OOM.

I'm going to try tuning some parameters to see if I can make that less likely.

PR #812, opened 2025-11-06
feat(config): add support for multiple named layouts

PaulHax:
Replace single layout config field with layouts object that supports multiple named layouts. Users can now define several layouts in their config and switch between them via the UI while preserving view data.

- Add layouts object to config schema (key-value pairs)
- Add namedLayouts state and methods to view store
- Create LayoutSelector component with named layouts list and grid editor
- Update ControlsStrip to use LayoutSelector
- Update e2e tests to use new layouts format
- Add test for multiple named layouts functionality

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -2,15 +2,15 @@
 
 By loading a JSON file, you can set VolView's configuration:
 
-- Starting view layout (grid size, view types, or hierarchical layouts)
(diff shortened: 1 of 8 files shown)


zachmullen:
@andinet FYI, layout selection now supports both old and new modes: 

Screenshot 2025-11-07 at 7 43 47 AM

andinet:
Great. thank you!

PR #839, opened 2026-01-10
Fix updateSegment function proxy the segments js object

1isten:
The `segmentGroupStore.updateSegment` function some how makes the `metadata.segments.byValue.color` field a proxy array.

Steps to reproduce: click the [Lock] or [Hide] toggle button, which calls the `updateSegment` function:
image

then, this object becomes proxied:

image

This leads to errors in below scenarios:

1. Failed to create new segment group, which calls the `structuredClone`
image
image

2. Failed to save session, which calls the `postMessage`
image
image


PaulHax:
Nice find.  Thank you again.

PR #879, opened 2026-05-17
chore(deps): upgrade @kitware/vtk.js to 36.1.1

PaulHax:
Bumps @kitware/vtk.js from `^32.12.1` to `^36.1.1`.

## API migration

Per vtk.js BREAKING_CHANGES (33 -> 34), several volume mapper setters were moved from the mapper onto the volume property. Callers are adjusted to invoke the new property-based setters:

- `src/utils/volumeProperties.ts`: route the relevant `set*` calls through the volume's property instead of the mapper.
- `src/components/vtk/VtkBaseVolumeRepresentation.vue`: same migration for the representation component's mapper/property interactions.

## Build config

vtk.js v36 builds with Vite/Rolldown (Kitware/vtk-js#3435). Its internal `ColorMaps.json.js` is now emitted with `export default` instead of v35's minified named export, so the `virtual-modules` stub in `vite.config.ts` is updated to match.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@itk-wasm/dicom": "^7.6.4",
         "@itk-wasm/image-io": "1.6.1",
         "@itk-wasm/morphological-contour-interpolation": "2.0.0",
-        "@kitware/vtk.js": "^32.12.1",
+        "@kitware/vtk.js": "^36.1.1",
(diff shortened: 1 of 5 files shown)


PaulHax:
done by https://github.com/Kitware/VolView/pull/895

